PR #12, opened 2020-04-23
Feature#11 + Bug fixes + More test cases

### LINKED ISSUE #11: VRF prove method returning also output hash
Considering that a single node can be delegated with several forger boxes we should optimize the lookup for an eligible forger box for a particular slot.
A possible optimization will be to directly return the VRF output hash during proof creation to avoid for each forger box to calculate the proof, validate it again and then extract the output hash to be able to verify eligibility considering the box relative stake.

DanieleDiBenedetto:
Closes #11 , corrects some bugs and expanded the Java tests


jni/src/main/java/com/horizen/vrfnative/VRFOutput.java
```diff
@@ -0,0 +1,27 @@
+package com.horizen.vrfnative;
+
+import com.horizen.librustsidechains.FieldElement;
+import com.horizen.librustsidechains.Library;
+
+
+public class VRFOutput {
```
albertog78, line 7: VRFOutput can be confused with the real output of the VRF. Let's call it VRFProveResult that is containing the vfrProof and the vrfOutput

jni/src/test/java/com/horizen/schnorrnative/SchnorrKeyPairTest.java
```diff
@@ -7,6 +7,7 @@
 
 import static org.junit.Assert.*;
 import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.assertFalse;
 
 public class SchnorrKeyPairTest {
```
albertog78, line 12: introduce some regression test using pre-computed data

jni/src/test/java/com/horizen/sigproofnative/NaiveThresholdSigProofTest.java
```diff
@@ -14,6 +14,7 @@
 
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.assertFalse;
 
 public class NaiveThresholdSigProofTest {
```
albertog78, line 19: add regression test

jni/src/main/java/com/horizen/librustsidechains/FieldElement.java
```diff
@@ -12,9 +14,57 @@ private FieldElement(long fieldElementPointer) {
...
+    private native byte[] nativeSerializeFieldElement();
+
+    public byte[] serializeFieldElement() {
+        if (fieldElementPointer == 0)
+            throw new IllegalArgumentException("Field element was freed.");
+
+        return nativeSerializeFieldElement();
+    }
+
+    private static native long nativeDeserializeFieldElement(byte[] fieldElementBytes);
```
SergeyMakukha, line 40: Currently we cannot deserialize field element basing on random bytes. I think we need have method to deserialize (or create new) field element from random (unpredictible) bytes.

api/src/lib.rs
```diff
@@ -1105,6 +1116,61 @@ pub extern "system" fn Java_com_horizen_sigproofnative_NaiveThresholdSigProof_na
...
+    _env.byte_array_from_slice(fe_bytes.as_ref())
+        .expect("Cannot write field element.")
+}
+
+#[no_mangle]
+pub extern "system" fn Java_com_horizen_librustsidechains_FieldElement_nativeDeserializeFieldElement(
+    _env: JNIEnv,
+    _class: JClass,
+    _field_element_bytes: jbyteArray,
+) -> *mut FieldElement
```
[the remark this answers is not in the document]
DanieleDiBenedetto: Yes, I noticed that and I agree. Which one is the most efficient according to you (or are they the same) ?

jni/src/main/java/com/horizen/librustsidechains/FieldElement.java
```diff
@@ -12,9 +14,57 @@ private FieldElement(long fieldElementPointer) {
...
+
+    public static FieldElement deserialize(byte[] fieldElementBytes) {
+        if (fieldElementBytes.length != FIELD_ELEMENT_LENGTH)
```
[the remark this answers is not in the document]
DanieleDiBenedetto: Is it really needed ? Because that would be a particular case, and these cases are already handled by JNI functions which have very specific input parameters (e.g. the createNaiveThresholdProof function that takes the block hashes as 32 byte array and then does the conversion with padding internally)

jni/src/main/java/com/horizen/librustsidechains/FieldElement.java
DanieleDiBenedetto: Mmm is that really needed ? Are there situations in which you need to get a field element from random bytes ? If it's only for testing purposes, you have the function FieldElement.createRandom() that should be enough. Have you in mind specific use cases ?

SergeyMakukha: Resolved. As was discussed today on meeting.

api/src/lib.rs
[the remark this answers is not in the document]
SergeyMakukha: I think we need to use second approach because in some cases (like for VRFProveResult) Rust function returns 2 pointers.

jni/src/main/java/com/horizen/librustsidechains/FieldElement.java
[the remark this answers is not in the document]
SergeyMakukha: Resolved.

api/src/lib.rs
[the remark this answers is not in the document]
DanieleDiBenedetto: Done

PR #27, opened 2020-06-07
Add travis pipeline publishing to maven central via oss.sonatype.org

cronicc:
* Add docker container with all build and gpg sign prerequisites
* Update pom.xml with metadata (groupId, artifactId, name, description, licenses, developers, scm, issueManagement sections)
* Add sonatype repository information
* Add build-extras profile generating sources.jar and javadoc.jar
* Add sign profile signing bundle components with infrastructure@zensystem.io gpg maven subkey
* Any tag signed by a release engineer in format $SEMVER-SNAPSHOT will be deployed to https://oss.sonatype.org/content/repositories/snapshots
* Any tag signed by a release engineer in format $SEMVER will be deployed to https://oss.sonatype.org/service/local/staging/deploy/maven2 and pushed to https://repo1.maven.org/maven2/io/horizen/zendoo-sc-cryptolib
* Git tag and pom.xml project.version have to match for deployment to work
* Java and Rust versions can be changed in .travis.yml


jni/pom.xml
```diff
@@ -1,57 +1,165 @@
...
-						</configuration>
-					</execution>
-				</executions>
-			</plugin>
-		</plugins>
-	</build>
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>io.horizen</groupId>
+  <artifactId>zendoo-sc-cryptolib</artifactId>
+  <version>0.2-SNAPSHOT</version>
```
i-Alex, line 6: We should avoid using `-SNAPSHOT` in the versions we want to publish. Let's name it `0.3.0` or even `1.0.0`. @albertog78  what do you think about this?

albertog78: The release procedure should automatically release it as 0.2 and the commit the new pom with version 0.3-SNAPSHOT. @cronicc @lander86

cronicc: @i-Alex @albertog78 

In the CI pipeline we check that git tag == pom.project.version. pom.project.version has to be a non SNAPSHOT version in order for the bundle to be published to maven central, any SNAPSHOT versions will be published to the sonatype snapshot repository at https://oss.sonatype.org/content/repositories/snapshots.

In any case publishing to maven central or the SNAPSHOT repo is only triggered if a git tag exists.

As a general rule doing a release should be it's own pull request, where the version number is bumped and separate review is being done, but lets skip this this time around. I'll set version to 0.3.0.

PR #33, opened 2020-06-25
New PoseidonHash/RAMT/SMT

### LINKED ISSUE #14: Poseidon Hash/RAMT/SMT
To be able to not have to collect first all the inputs together and then passing them to poseidon, expose via JNI the new implementation coming from ginger lib (https://github.com/ZencashOfficial/ginger-lib/issues/56)

DanieleDiBenedetto:
Closes #14 and also #39


api/Cargo.toml
```diff
@@ -7,9 +7,9 @@ edition = "2018"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-algebra = {version = "0.1.0", git = "https://github.com/ZencashOfficial/ginger-lib.git", branch = "development"}
-primitives = {version = "0.1.0", features = ["merkle_tree", "signature", "vrf"], git = "https://github.com/ZencashOfficial/ginger-lib.git", branch = "development"}
-proof-systems = {version = "0.1.0", features = ["groth16"], git = "https://github.com/ZencashOfficial/ginger-lib.git", branch = "development"}
+algebra = {version = "0.1.0", git = "https://github.com/ZencashOfficial/ginger-lib.git", branch = "updatable_poseidon"}
```
DanieleDiBenedetto, line 10: Once UpdatablePoseidon branch will be merged in GingerLib we must change this dependency.

demo-circuit/Cargo.toml
```diff
@@ -6,12 +6,12 @@ edition = "2018"
 include = ["Cargo.toml", "src"]
 
 [dependencies]
-algebra = {version = "0.1.0", git = "https://github.com/ZencashOfficial/ginger-lib.git", branch = "development"}
-primitives = {version = "0.1.0", features = ["signature"], git = "https://github.com/ZencashOfficial/ginger-lib.git", branch = "development"}
-r1cs-crypto = {version = "0.1.0", features = ["signature"], git = "https://github.com/ZencashOfficial/ginger-lib.git", branch = "development"}
-r1cs-core = {version = "0.1.0", git = "https://github.com/ZencashOfficial/ginger-lib.git", branch = "development"}
-r1cs-std = {version = "0.1.0", git = "https://github.com/ZencashOfficial/ginger-lib.git", branch = "development"}
-proof-systems = {version = "0.1.0", features = ["groth16"], git = "https://github.com/ZencashOfficial/ginger-lib.git", branch = "development"}
+algebra = {version = "0.1.0", git = "https://github.com/ZencashOfficial/ginger-lib.git", branch = "updatable_poseidon"}
```
DanieleDiBenedetto, line 9: Once UpdatablePoseidon branch will be merged in GingerLib we must change this dependency.

ouroboros/Cargo.toml
```diff
@@ -5,7 +5,7 @@ authors = ["DanieleDiBenedetto <daniele@horizenlabs.io>"]
 edition = "2018"
 
 [dependencies]
-algebra = {version = "0.1.0", git = "https://github.com/ZencashOfficial/ginger-lib.git", branch = "development"}
-primitives = {version = "0.1.0", features = ["vrf"], git = "https://github.com/ZencashOfficial/ginger-lib.git", branch = "development"}
+algebra = {version = "0.1.0", git = "https://github.com/ZencashOfficial/ginger-lib.git", branch = "updatable_poseidon"}
```
DanieleDiBenedetto, line 8: Once UpdatablePoseidon branch will be merged in GingerLib we must change this dependency.

api/src/ginger_calls.rs
```diff
@@ -169,9 +178,25 @@ pub fn read_field_element_from_u64(num: u64) -> FieldElement {
 // in MC during SC creation.
 pub fn compute_pks_threshold_hash(pks: &[SchnorrPk], threshold: u64) -> Result<FieldElement, Error> {
     let threshold_field = read_field_element_from_u64(threshold);
-    let pks_x = pks.iter().map(|pk| pk.x).collect::<Vec<_>>();
-    let pks_hash = compute_poseidon_hash(pks_x.as_slice())?;
-    compute_poseidon_hash(&[pks_hash, threshold_field])
+    let mut h = FieldHash::init(None);
+    pks.iter().for_each(|pk| { h.update(pk.x); });
+    let pks_hash = h.finalize();
+    Ok(h
```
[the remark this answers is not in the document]
DanieleDiBenedetto: Return type is outdated, because it's considering previous hash function. I will fix it.

api/src/ginger_calls.rs
```diff
@@ -169,9 +178,25 @@ pub fn read_field_element_from_u64(num: u64) -> FieldElement {
...
+        .reset(None)
+        .update(pks_hash)
+        .update(threshold_field)
+        .finalize())
+}
+
+const BT_MERKLE_TREE_HEIGHT: usize = 13;
+
+fn compute_bt_root(bts: &[FieldElement]) -> Result<FieldElement, Error> {
+    let mut bt_mt = GingerRAMT::init(2usize.pow((BT_MERKLE_TREE_HEIGHT - 1) as u32));
```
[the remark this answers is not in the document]
DanieleDiBenedetto: It's the same, here I'm directly calling the GingerLib function and not the Cryptolib wrappers. For internal function implementations I think it's acceptable.

api/src/ginger_calls.rs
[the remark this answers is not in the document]
DanieleDiBenedetto: Done

jni/src/main/java/com/horizen/merkletreenative/BigLazyMerkleTree.java
```diff
@@ -28,15 +30,15 @@ public static BigLazyMerkleTree init(int height, String statePath, String dbPath
     // Returns the position to which insert the leaf
     public long getPosition(FieldElement leaf) {
         if (lazyMerkleTreePointer == 0)
-            throw new IllegalArgumentException("lazyMerkleTreePointer must be not null.");
+            throw new IllegalArgumentException("LazyBigMerkleTree instance was freed.");
```
i-Alex, line 33: I would suggest to change `IllegalArgumentException` to `IllegalStateException` type. Same bellow.

jni/src/main/java/com/horizen/merkletreenative/BigMerkleTree.java
```diff
@@ -26,42 +28,50 @@ public static BigMerkleTree init(int height, String statePath, String dbPath, St
...
     public boolean isPositionEmpty(long position){
         if (merkleTreePointer == 0)
-            throw new IllegalArgumentException("merkleTreePointer must be not null.");
+            throw new IllegalArgumentException("BigMerkleTree instance was freed.");
         return nativeIsPositionEmpty(position);
     }
 
     private native void nativeAddLeaf(FieldElement leaf, long position);
 
     public void addLeaf(FieldElement leaf, long position){
```
i-Alex, line 45: So, what about this method? Are you going to remove `position` arg?

jni/src/main/java/com/horizen/merkletreenative/InMemoryOptimizedMerkleTree.java
```diff
@@ -0,0 +1,106 @@
...
+    // just once, but all the leaves must be kept in memory).
+    public static InMemoryOptimizedMerkleTree init(int height, long processingStep){
+        return nativeInit(height, processingStep);
+    }
+
+    private native void nativeAppend(FieldElement input);
+
+    public void append(FieldElement input) {
+        if (inMemoryOptimizedMerkleTreePointer == 0)
+            throw new IllegalArgumentException("InMemoryOptimizedMerkleTree instance was freed.");
```
i-Alex, line 42: -> `IllegalStateException`

jni/src/main/java/com/horizen/merkletreenative/MerklePath.java
```diff
@@ -0,0 +1,41 @@
...
+        Library.load();
+    }
+
+    private MerklePath(long merklePathPointer) {
+        if (merklePathPointer == 0)
+            throw new IllegalArgumentException("merklePathPointer must be not null.");
+        this.merklePathPointer = merklePathPointer;
+    }
+
+    private native boolean nativeVerify(int merkleTreeHeight, FieldElement leaf, FieldElement root);
```
i-Alex, line 20: Why do we need the `merkleTreeHeight` here? Is it not enough just to apply the path elements to the leaf one by one and then compare it with the `root`?

i-Alex (changes requested):

jni/src/main/java/com/horizen/merkletreenative/BigMerkleTree.java
DanieleDiBenedetto: I'm not sure about this one. From an usage perspective can be misleading, but probably for testing and debugging it will be useful

jni/src/main/java/com/horizen/merkletreenative/MerklePath.java
DanieleDiBenedetto: I want to check also that the path is of the correct length, otherwise I can just pass as a leaf a children of the root and the method will pass anyway.
As using the height as an explicit parameter, instead, is a result of keeping the height as a free parameter. Hardcoding the height in here means that if we want to modify the height of a tree, we should release a new version of the cryptolib

albertog78 (changes requested):

Issue #31, opened 2020-06-17
Add exception handling to JNI and Java side


DanieleDiBenedetto:
At the moment, when we have to deal with exceptional cases without crashing, we return JNI_FALSE or NULL, but this doesn't really help to understand what is the cause of the exception.
Add the possibility to throw Exceptions from Rust-JNI and handle them from Java side.

Issue #22, opened 2020-05-21
Clean up and uniform JNI code


DanieleDiBenedetto:
We should refactor the JNI code in api/src/lib to make it cleaner:

- Remove all code duplications (e.g. when creating a FieldElement jobject to return);
- Remove all unnecessary input parameters (like the JClass when they are not needed);
- Uniform the logic to perform certain operations (e.g. the "free" functions are sometimes implemented by taking as argument the pointer to free, other times by directly working with the JObject containing the pointer).

DanieleDiBenedetto:
For what regards:

`Java classes and methods can be cached at startup (e.g. by implementing JNI_onLoad() function and putting the code inside it), so that it won't be necessary to perform a lookup each time they are needed (like in https://github.com/exonum/exonum-java-binding/tree/master/exonum-java-binding/core/rust);`

I benched it locally and it seems to not lead to any significant benefit, so let's give to this point a low priority.

PR #43, opened 2020-12-11
Bitvector

nastenko:
Initial implementation of the Bitvector functionality backed by BigMerkleTree

### DIFF
--- a/api/src/ginger_calls.rs
+++ b/api/src/ginger_calls.rs
@@ -552,7 +552,7 @@ pub fn get_ginger_smt(height: usize, db_path: &str) -> Result<GingerSMT, Error>{
     }
 }
 
-fn new_ginger_smt(height: usize, db_path: &str) -> Result<GingerSMT, Error> {
+pub fn new_ginger_smt(height: usize, db_path: &str) -> Result<GingerSMT, Error> {
     match GingerSMT::new(
         height,
         true,
@@ -563,7 +563,7 @@ fn new_ginger_smt(height: usize, db_path: &str) -> Result<GingerSMT, Error> {
     }
 }
 
-fn restore_ginger_smt(height: usize, db_path: &str) -> Result<GingerSMT, Error>
+pub fn restore_ginger_smt(height: usize, db_path: &str) -> Result<GingerSMT, Error>
 {
     match GingerSMT::load_batch::<GingerMerkleTreeParameters>(
         height,
@@ -598,6 +598,10 @@ pub fn add_leaf_to_ginger_smt(tree: &mut GingerSMT, leaf: &FieldElement, positio
 
 pub fn remove_leaf_from_ginger_smt(tree: &mut GingerSMT, position: u64){
     tree.remove_leaf(Coord::new(0, position as usize));
+}
+
+pub fn get_leaf_from_ginger_smt(tree: &GingerSMT, position: u64) -> Option<FieldElement> {
+    tree.get_leaf(Coord::new(0, position as usize))
 }
 
 pub fn get_ginger_smt_root(tree: &GingerSMT) -> FieldElement {
@@ -685,92 +689,6 @@ pub fn get_lazy_ginger_smt_path(tree: &mut LazyGingerSMT, leaf_position: u64) ->
     tree.get_merkle_path(Coord::new(0, leaf_position as usize))
 }
 
-////////////BITVECTOR BASED ON SPARSE MERKLE TREE
-
-pub type BitvectorTree = BigMerkleTree<GingerMerkleTreeParameters>;
-
-// Size of a BVT leaf in bits
-pub const BVT_LEAF_SIZE: u64 = FqParameters::CAPACITY as u64;
-
-pub fn get_bvt(height: usize, db_path: &str) -> Result<BitvectorTree, Error>{
-    // If at least the leaves database is available, we can restore the tree
-    if Path::new(db_path).exists() {
-        restore_ginger_smt(height, db_path)
-    } else { // Otherwise we need to create a new tree
-        new_ginger_smt(height, db_path)
-    }
-}
-
-pub fn flush_bvt(tree: &mut BitvectorTree) {
-    tree.flush()
-}
-
-pub fn set_bvt_persistency(tree: &mut BitvectorTree, persistency: bool) {
-    tree.set_persistency(persistency)
-}
-
-pub fn get_bvt_root(tree: &BitvectorTree) -> FieldElement {
-    tree.get_root()
-}
-
-pub fn get_bvt_path(tree: &mut BitvectorTree, bit_position: u64) -> GingerMHTPath {
-    tree.get_merkle_path(Coord::new(0, (bit_position / BVT_LEAF_SIZE) as usize))
-}
-
-pub fn get_bvt_leaf(tree: &BitvectorTree, bit_position: u64) -> Option<FieldElement> {
-    get_bvt_leaf_by_index(tree, bit_position / BVT_LEAF_SIZE)
-}
-
-pub fn set_bvt_bit(tree: &mut BitvectorTree, bit_position: u64){
-    modify_bit_in_bvt(tree, bit_position, true)
-}
-
-pub fn reset_bvt_bit(tree: &mut BitvectorTree, bit_position: u64){
-    modify_bit_in_bvt(tree, bit_position, false)
-}
-
-pub fn get_bvt_bit(tree: &BitvectorTree, bit_position: u64) -> bool {
-    if let Some(leaf) = get_bvt_leaf(tree, bit_position) {
-        let leaf_bits= leaf.write_bits();
-        assert_eq!(BVT_LEAF_SIZE, (leaf_bits.len() - 1) as u64);
-        // The LSB bits are at the MSB positions
-        leaf_bits[(BVT_LEAF_SIZE - (bit_position % BVT_LEAF_SIZE)) as usize]
-    } else { // leaf doesn't even exist, so the bit isn't set
-        false
-    }
-}
-
-fn get_bvt_leaf_by_index(tree: &BitvectorTree, leaf_index: u64) -> Option<FieldElement> {
-    tree.get_leaf(Coord::new(0, leaf_index as usize))
-}
-
-fn modify_bit_in_bvt(tree: &mut BitvectorTree, bit_position: u64, bit_value: bool){
-    use algebra::{Field, FromBits};
-
-    let updated_leaf = FieldElement::read_bits(
-        {
-            let mut leaf_bits =
-                if let Some(leaf) = get_bvt_leaf(tree, bit_position) {
-                    leaf.write_bits()
-                } else { // leaf doesn't exist, so create an empty leaf
-                    FieldElement::zero().write_bits()
-                };
-            assert_eq!(BVT_LEAF_SIZE, (leaf_bits.len() - 1) as u64);
-            // The LSB bits are at the MSB positions
-            leaf_bits[(BVT_LEAF_SIZE - (bit_position % BVT_LEAF_SIZE)) as usize] = bit_value;
-            leaf_bits
-        }
-    ).unwrap();
-
-    let leaf_index = (bit_position / BVT_LEAF_SIZE) as usize;
-    if !updated_leaf.is_zero() {
-        // Write an updated leaf value
-        tree.insert_leaf(Coord::new(0, leaf_index), updated_leaf);
-    } else {
-        // Don't store empty leafs to optimize SMT size
-        tree.remove_leaf(Coord::new(0, leaf_index))
-    }
-}
 
 #[cfg(test)]
 mod test {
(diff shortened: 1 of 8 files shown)


api/src/ginger_calls.rs
```diff
@@ -684,6 +684,95 @@ pub fn get_lazy_ginger_smt_path(tree: &mut LazyGingerSMT, leaf_position: u64) ->
...
+    // println!("updated_leaf: {}", updated_leaf);
+    if !updated_leaf.is_zero() {
+        // Write an updated leaf value
+        add_leaf_to_ginger_smt(tree, &updated_leaf, leaf_index)
+    } else {
+        // Don't store empty leafs to optimize SMT size
+        remove_leaf_from_ginger_smt(tree, leaf_index)
+    }
+}
+
```
DanieleDiBenedetto, line 775: I think it's still needed a function that returns the whole bitvector, something like:

fn get_bitvector(tree: &BitVectorTree) -> Vec\<bool\>

Probably there might be troubles implementing this function with the implementation you provided:
1) You need a way to keep track of the biggest bit index so that you know where to stop when serializing the last leaf;
2) If for some reason you unset one or more bits of a leaf, and all the bits of that leaf will be unset, then in line 772 you are going to remove that leaf and you won't be able anymore to return it (but it should be present because that utxos exist). There shouldn't be any reason to unset a bit of the bitvector, except at each epoch start, when you reset all the bits of the bitvector (but I think you won't do it with this function), so maybe this is a false problem.

DanieleDiBenedetto, line 775: In fact, in relation to the previous comment, how do you plan to unset all the bits of the bitvector ?

DanieleDiBenedetto (changes requested):

api/src/ginger_calls.rs
```diff
@@ -684,6 +684,95 @@ pub fn get_lazy_ginger_smt_path(tree: &mut LazyGingerSMT, leaf_position: u64) ->
...
+        false
+    }
+}
+
+fn bits_per_bvt_leaf() -> u64 {
+    (FieldElement::size_in_bits() - 1) as u64 // using one bit lesser to make leafs never exceed the field modulus
+}
+
+fn get_leaf_in_bvt(tree: &BitvectorTree, leaf_position: u64) -> Option<FieldElement> {
+    tree.get_leaf(Coord::new(0, leaf_position as usize))
```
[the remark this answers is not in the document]
nastenko: Yes, this would be an easier approach, but as I understood from implementation of the insert_leaf method in BMT, there is a check for presence of a specified node before retrieving it from database:
 if self.state.present_node.contains(&coord) {
            let old_hash = self.get_from_db(coord.idx).unwrap();

https://github.com/HorizenOfficial/ginger-lib/blob/0dc2e545aa5b1b9ce291c713ad47f99f654bc8ca/primitives/src/merkle_tree/field_based_mht/smt/big_merkle_tree.rs#L411

So to be consistent with BMT implementation I proposed additional method with this check and further fetching from db. But if this check for presence is not necessarry and direct access to DB in case if node is not actually present will not cause unexpected errors then we can just directly use the get_from_db method.

api/src/ginger_calls.rs
nastenko: Regarding deletion of the leafs when they don't contain any set bit: I mean that the absent leaf means that the corresponding chunk of bits is all zero. Initially the whole BVT consists of emty leafs that corresponds to all zero bits in BVT. So there is no need to store zeroized leafs. Also this will reduce size of the underlying DB. 
But actulally I also don't see the scenario where some bit should be reset except for the case, when some bit was set unintentionally and need to be unset without re-initializing whole BVT.

About serialization of the BVT - I think it is enough to get all non-empty leafs together with their positions, and then put them back in the same positions during deserialization. All empty leafs are assumed "to contain" only zero-bits, so no need to serialize them, because we already know that all empty leafs "contain" zeros.

And regarding the fn get_bitvector(tree: &BitVectorTree) -> Vec<bool>: I think it may be not possible to return all bits from BVT because based of 64-bit size of bits indexes there are at maximum 2^64 bits, so the vector to be returned is 2^61 Bytes = 2^31 GB in size, that is too much. But for a smaller trees it can be feasible. 
But actually any bit can be retrieved just by calling get_bvt_bit. So the need for get_bitvector method is a bit questionable from my viewpoint.

nastenko: Based on the assumption that empty leaf implies all-zero vector, it is enough to delete all leafs to unset a whole BVT

api/src/ginger_calls.rs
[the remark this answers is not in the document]
DanieleDiBenedetto: Yes, because the present_node is stored in memory, so I guess it's a quick way to avoid looking on disk if you already know that the leaf is not present. Okey, let's proceed with the get_leaf() function !

api/src/ginger_calls.rs
DanieleDiBenedetto: It's not clear to me what will be the usage of the whole bitvector, if needed at all. But in case it's needed, I think we must serialize also the zero leaves, because they correspond to actual existing UTXOs (plus an empty leaf can be in the middle of two non empty ones). Maybe we can ask @i-Alex about this

api/src/zenbox_smt/bitvector_smt.rs
```diff
@@ -0,0 +1,192 @@
...
+use primitives::{BigMerkleTree, Coord};
+use crate::ginger_calls::{GingerMerkleTreeParameters, Error, FieldElement, GingerMHTPath, GingerSMT, restore_ginger_smt, new_ginger_smt};
+use algebra::fields::mnt6753::FqParameters;
+use algebra::{FpParameters, ToBits};
+use std::path::Path;
+
+pub type BitvectorSMT = BigMerkleTree<GingerMerkleTreeParameters>;
+
+// Computes floor(log2(x))
+const fn log2(x: u64) -> usize {
```
DanieleDiBenedetto, line 12: We already have this function in GingerLib, in algebra/mod.rs

api/src/zenbox_smt/mod.rs
```diff
@@ -0,0 +1,226 @@
...
+pub fn is_box_spent(tree: &ZenBoxSMT, position: u64) -> bool {
+    get_bvt_bit(&tree.bitvector_smt, position)
+}
+
+pub fn add_box_to_zenbox_smt(tree: &mut ZenBoxSMT, box_: &FieldElement, position: u64){
+    // avoid overwriting already existing boxes
+    if is_position_empty_in_ginger_smt(&tree.state_smt, position){
+        // if corresponding bit in BVT is set, this means that current box has already been spent
+        assert!(!get_bvt_bit(&tree.bitvector_smt, position));
+        add_leaf_to_ginger_smt(&mut tree.state_smt, box_, position)
```
DanieleDiBenedetto, line 71: Shouldn't we set the corresponding bit in the bvt also in the case of new created boxes ?

DanieleDiBenedetto (commented):
In general ok, a couple of thoughts:
1) When we are going to do the bit vector compression/decompression, I think you will need to keep track of additional info, but we will see when the time comes;
2) I think JNI side both BitVectorSMT and ZenBoxSMT can extend the BigMerkleTree class, so you will be able to re-use some of the already existing functions and avoid some code duplication;
3) According to the logic SDK side (e.g. if we don't need to update the trees one box at a time), it may be worth to use a LazyBigMerkleTree instead of a BigMerkleTree for performance improvement.

api/src/zenbox_smt/bitvector_smt.rs
nastenko: Thanks, then I will use it from Ginger!

api/src/zenbox_smt/mod.rs
nastenko: As I understand the BVT functionality, it should track the Boxes that were spent - 1 or never spent - 0. But I will clarify this.

nastenko:
> In general ok, a couple of thoughts:
> 
> 1. When we are going to do the bit vector compression/decompression, I think you will need to keep track of additional info, but we will see when the time comes;
> 2. I think JNI side both BitVectorSMT and ZenBoxSMT can extend the BigMerkleTree class, so you will be able to re-use some of the already existing functions and avoid some code duplication;
> 3. According to the logic SDK side (e.g. if we don't need to update the trees one box at a time), it may be worth to use a LazyBigMerkleTree instead of a BigMerkleTree for performance improvement.

Thank you for the additional suggestions!
Regarding the LazyBigMerkleTree - indeed, we can implement additional versions such as LazyZenBoxSMT and LazyBitVectorSMT like it is done for BigMerkleTree and LazyBigMerkleTree.

About the JNI side I suppose to create class ZenBoxSMT with JNI calls to the rust methods of ZenBoxSMT, like it is done for BigMerkleTree. Maybe I misunderstood, but I don't see how BitVectorSMT and ZenBoxSMT can be extended from BigMerkleTree class on the Java side.

DanieleDiBenedetto:
> Thank you for the additional suggestions!
>
Thanks to you !
> Regarding the LazyBigMerkleTree - indeed, we can implement additional versions such as LazyZenBoxSMT and LazyBitVectorSMT like it is done for BigMerkleTree and LazyBigMerkleTree.
>
Only if it's worth it: if the logic SDK-side is such that you have to update the tree with lot of boxes in a single time, the Lazy will be much quicker; otherwise we can keep the non-lazy.
> About the JNI side I suppose to create class ZenBoxSMT with JNI calls to the rust methods of ZenBoxSMT, like it is done for BigMerkleTree. Maybe I misunderstood, but I don't see how BitVectorSMT and ZenBoxSMT can be extended from BigMerkleTree class on the Java side.
>
You can create, for instance, a BitVectorSMT class in JNI folder extending BigMerkleTree (or LazyBigMerkleTree according to the comment above): in this way you already have the function to create, destroy, get root, get path, etc.; then you can add in this new class the specific BitVector functions such as modify_bit_in_bvt, get_bvt_bit and so on.

nastenko:
> You can create, for instance, a BitVectorSMT class in JNI folder extending BigMerkleTree (or LazyBigMerkleTree according to the comment above): in this way you already have the function to create, destroy, get root, get path, etc.; then you can add in this new class the specific BitVector functions such as modify_bit_in_bvt, get_bvt_bit and so on.

Thank you, now I understood! Actually this is the way to implement ZenBoxSMT which I also thought about, but Alberto said that it would be better to implement the whole logic completely in Rust. As I understand this will save JNI-calls, because in case of extending BigMerkleTree on the Java side there would be at least two JNI calls for methods which work with BitVectorSMT: to the BigMerkleTree and to BitVectorSMT both on the Rust side. But in case when on the Rust side we have ZenBoxSMT, which contains BigMerkleTree and BitVectorSMT and exposes methods to the Java side there will be a single JNI call per each method, while all the auxiliary logic for updating both trees is implemented in Rust.

DanieleDiBenedetto:
> > You can create, for instance, a BitVectorSMT class in JNI folder extending BigMerkleTree (or LazyBigMerkleTree according to the comment above): in this way you already have the function to create, destroy, get root, get path, etc.; then you can add in this new class the specific BitVector functions such as modify_bit_in_bvt, get_bvt_bit and so on.
> 
> Thank you, now I understood! Actually this is the way to implement ZenBoxSMT which I also thought about, but Alberto said that it would be better to implement the whole logic completely in Rust. As I understand this will save JNI-calls, because in case of extending BigMerkleTree on the Java side there would be at least two JNI calls for methods which work with BitVectorSMT: to the BigMerkleTree and to BitVectorSMT both on the Rust side. But in case when on the Rust side we have ZenBoxSMT, which contains BigMerkleTree and BitVectorSMT and exposes methods to the Java side there will be a single JNI call per each method, while all the auxiliary logic for updating both trees is implemented in Rust.

Maybe we are saying the same thing, let's proceed with your design, mine was just a minor comment !

jni/src/main/java/com/horizen/merkletreenative/ZenBoxMerkleTree.java
```diff
@@ -0,0 +1,159 @@
...
+    private native FieldElement nativeBitvectorRoot();
+    private native MerklePath nativeGetStateMerklePath(long leafPosition);
+    private native MerklePath nativeGetBitvectorMerklePath(long leafPosition);
+    private native void nativeFlush();
+
+    // Creates ZenBoxSMT Rust-instance, which though can be not created in case of underlying errors
+    // The returned ZenBoxMerkleTree object should be checked with the 'isInitialized' method
+    // Note: ZenBoxSMT Rust-instance may be not created due to errors such as incorrectly specified tree height or DB path, insufficient disk space and so on.
+    //       In such a case constructor will be called with merkleTreePointer = null so ZenBoxMerkleTree will be empty.
+    //       Thus ZenBoxMerkleTree should be checked with 'isInitialized' method before any usage
```
DanieleDiBenedetto, line 48: Nice, but in the future this won't be the way we deal with it.
Look at this opened issue:

https://github.com/HorizenOfficial/zendoo-sc-cryptolib/issues/31

jni/src/main/java/com/horizen/merkletreenative/ZenBoxMerkleTree.java
```diff
@@ -0,0 +1,159 @@
...
+
+    // Removes a Box from a specified position of the State Merkle Tree
+    // Also sets a bit correspondingly to the specified position in the Bitvector Merkle Tree
+    public void removeBox(long boxPosition){
+        checkPointer();
+        nativeRemoveBox(boxPosition);
+    }
+
+    // Returns FieldElement which can be empty if a Box at a specified position doesn't exist
+    // Should be checked with the FieldElement's 'nonEmpty' method
```
DanieleDiBenedetto, line 117: Same reasoning as comment above

api/src/zenbox_smt/jni_calls.rs
```diff
@@ -0,0 +1,308 @@
...
+}
+
+fn unwrap_ptr<'a, T: 'static>(env: &JNIEnv, ptr: JObject) -> &'a T {
+    read_raw_pointer(get_raw_ptr(env, ptr))
+}
+
+fn unwrap_mut_ptr<'a, T: 'static>(env: &JNIEnv, ptr: JObject) -> &'a mut T {
+    read_mut_raw_pointer(get_raw_ptr(env, ptr))
+}
+
```
DanieleDiBenedetto, line 35: Nice way to generalize, and it will be good if you apply it to all the JNI functions we have.
Of course this goes outside the scope of your task, but there's an issue opened about this:

https://github.com/HorizenOfficial/zendoo-sc-cryptolib/issues/22

We will do this kind of refactor, along with other refactoring and optimizations, when someone will deal with this issue. My suggestion is, for the moment, remove the functions above and use the ones already defined in api/lib.rs (just for the sake of uniformity), but save them somewhere for later

nastenko: Thanks Daniele! No problem, in my spare time I a can do at list partially such a refactoring by using these methods in lib.rs.

jni/src/main/java/com/horizen/merkletreenative/ZenBoxMerkleTree.java
nastenko: Yes, indeed it would be better to use exceptions to provide a more informative report about what happened inside of the Rust methods in case of some runtime errors

api/src/zenbox_smt/bitvector_smt.rs
```diff
@@ -0,0 +1,192 @@
+////////////BITVECTOR SPARSE MERKLE TREE
+
+use primitives::{BigMerkleTree, Coord};
+use crate::ginger_calls::{GingerMerkleTreeParameters, Error, FieldElement, GingerMHTPath, GingerSMT, restore_ginger_smt, new_ginger_smt};
+use algebra::fields::mnt6753::FqParameters;
+use algebra::{FpParameters, ToBits};
+use std::path::Path;
+
+pub type BitvectorSMT = BigMerkleTree<GingerMerkleTreeParameters>;
```
albertog78, line 9: @nastenko 
Introduce a function to serialize the Bitvector leaves to a byte array. The serialization should avoid to include the fieldelements bits not available in leaves. 
For example, with field modulus size 255 bits, the available bits are 254 and serialized size must be 254 bits and not 256 bits. This means that if we have to serialize 8 field elements the serialize size in bytes must 254 bytes instead of 256.

PR #44, opened 2021-03-25
Updatable poseidon tweedle

phoinic:


api/src/ginger_calls.rs
```diff
@@ -1,85 +1,117 @@
 use algebra::{fields::{
-    mnt4753::{Fr, Fq as ScalarFieldElement}, PrimeField
+    tweedle::{Fq as Fr, Fr as ScalarFieldElement}, PrimeField
```
DanieleDiBenedetto, line 2: Mmm you should sync your concrete types with the one used in zendoo-mc-cryptolib and zendoo-cctp-lib: I fear that in both of them tweedle::Fr is used as FieldElement and not Fq as you did

DanieleDiBenedetto (changes requested):

api/src/ginger_calls.rs
DanieleDiBenedetto: If all the types are not synced, the integration tests between MC and SC will never pass (they will read different fields)

phoinic: I have changed the FieldElement from Fq to Fr

api/src/ginger_calls.rs
```diff
@@ -1,85 +1,117 @@
...
-pub const SCALAR_FIELD_SIZE: usize = FIELD_SIZE;// 96
+// #[derive(Clone)]
+// struct MarlinNoLCNoZk;
+//
+// impl MarlinConfig for MarlinNoLCNoZk {
+//     const LC_OPT: bool = false;
+//     const ZK: bool = false;
+// }
+
+type IPAPC = InnerProductArgPC<PCAffine, Blake2s>;
```
DanieleDiBenedetto, line 68: For types and constants would be nice to do something similar to https://github.com/HorizenOfficial/zendoo-mc-cryptolib/blob/as/ginger_algebra_ocl/src/type_mapping.rs

api/src/ginger_calls.rs
```diff
@@ -215,7 +247,7 @@ pub fn compute_msg_to_sign(
 ) -> Result<(FieldElement, FieldElement), Error> {
 
     let mr_bt = if bt_list.is_empty() {
-        PHANTOM_MERKLE_ROOT //Note: Can actually be taken from MHT_PARAMETERS::nodes[BT_MERKLE_TREE_HEIGHT]
+        MHT_PARAMETERS.nodes[BT_MERKLE_TREE_HEIGHT].clone()
```
DanieleDiBenedetto, line 250: Ignore the TODO, let's stick with the PHANTOM_MERKLE_ROOT for the moment (if you see in Ginger it's generated starting from a magic string)

api/src/ginger_calls.rs
```diff
@@ -1,85 +1,117 @@
...
-pub const SCHNORR_SK_SIZE: usize = SCALAR_FIELD_SIZE; // 96
+pub const SCHNORR_SK_SIZE: usize = SCALAR_FIELD_SIZE; // 32
 pub const SCHNORR_SIG_SIZE: usize = 2 * FIELD_SIZE; // 192
 
 pub const VRF_PK_SIZE: usize = G1_SIZE; // 193
-pub const VRF_SK_SIZE: usize = SCALAR_FIELD_SIZE; // 96
+pub const VRF_SK_SIZE: usize = SCALAR_FIELD_SIZE; // 32
 pub const VRF_PROOF_SIZE: usize = G1_SIZE + 2 * FIELD_SIZE; // 192
 
 pub const ZK_PROOF_SIZE: usize = 2 * G1_SIZE + G2_SIZE;  // 771
```
DanieleDiBenedetto, line 85: This is the size of a Groth proof, we should change to the size of a standard Marlin one

api/src/ginger_calls.rs
```diff
@@ -306,22 +339,28 @@ pub fn create_naive_threshold_sig_proof(
...
+    let proof = MarlinInst::prove::<_, OsRng>(&pk, c, &mut rng).unwrap();
+
     Ok((proof, valid_signatures))
 }
 
 pub fn verify_naive_threshold_sig_proof(
     constant:                 &FieldElement,
-    end_epoch_mc_b_hash:      &[u8; 32],
-    prev_end_epoch_mc_b_hash: &[u8; 32],
+    end_epoch_mc_b_hash:      &[u8; 16],
```
DanieleDiBenedetto, line 357: These sizes are still 32: it's a MC hash, i.e. a SHA-256, they have nothing to do with our fields

api/src/ginger_calls.rs
```diff
@@ -716,8 +761,8 @@ mod test {
         let mut rng = OsRng;
 
         //Generate random mc block hashes and bt list
-        let mut end_epoch_mc_b_hash = [0u8; 32];
-        let mut prev_end_epoch_mc_b_hash = [0u8; 32];
+        let mut end_epoch_mc_b_hash = [0u8; 16];
```
DanieleDiBenedetto, line 764: Again: those are still 32

api/src/lib.rs
```diff
@@ -1269,58 +1307,58 @@ pub extern "system" fn Java_com_horizen_merkletreenative_InMemoryOptimizedMerkle
...
+//             .expect("Should be able to get field merkleTreePointer");
+//
+//         read_raw_pointer(t.j().unwrap() as *const GingerSMT)
+//     };
+//
+//     get_position_in_ginger_smt(tree, leaf) as jlong
+// }
 
 #[no_mangle]
 pub extern "system" fn Java_com_horizen_merkletreenative_BigMerkleTree_nativeGetAbsolutePosition(
```
DanieleDiBenedetto, line 1364: This is not using smt but I think it's not needed too

api/src/lib.rs
```diff
@@ -1341,218 +1379,218 @@ pub extern "system" fn Java_com_horizen_merkletreenative_BigMerkleTree_nativeGet
...
+//             .expect("Should be able to get field lazyMerkleTreePointer");
+//
+//         read_raw_pointer(t.j().unwrap() as *const LazyGingerSMT)
+//     };
+//
+//     get_position_in_lazy_ginger_smt(tree, leaf) as jlong
+// }
 
 #[no_mangle]
 pub extern "system" fn Java_com_horizen_merkletreenative_BigLazyMerkleTree_nativeGetAbsolutePosition(
```
DanieleDiBenedetto, line 1596: Same as above

api/src/lib.rs
```diff
@@ -2296,7 +2356,7 @@ pub extern "system" fn Java_com_horizen_sigproofnative_NaiveThresholdSigProof_na
         let t = _env.convert_byte_array(_end_epoch_block_hash)
             .expect("Should be able to convert to Rust array");
 
-        let mut end_epoch_block_hash_bytes = [0u8; 32];
+        let mut end_epoch_block_hash_bytes = [0u8; 16];
```
DanieleDiBenedetto, line 2359: As above: those are 32

api/src/lib.rs
```diff
@@ -2416,7 +2478,7 @@ pub extern "system" fn Java_com_horizen_sigproofnative_NaiveThresholdSigProof_na
...
```
DanieleDiBenedetto, line 2481: 32

demo-circuit/src/naive_threshold_sig/mod.rs
```diff
@@ -267,56 +291,57 @@ pub fn generate_parameters(max_pks: usize) -> Result<Parameters<MNT4>, Synthesis
         _field:                   PhantomData
     };
 
-    let params = generate_random_parameters::<MNT4, _, _>(c, &mut rng);
-    params
+    let universal_srs = MarlinInst::universal_setup(2usize.pow(20), 2usize.pow(20), 2usize.pow(20), &mut rng).unwrap();
```
DanieleDiBenedetto, line 294: If you want to give exact numbers here and not waste too much time in the generation procedure (you know that we are going to remove it by passing the segment size), the Marlin::index() function accepts only the circuit and in that function there are a couple of prints you can use to get exact data about the number of constraints, number of variables, and number of non zero

demo-circuit/src/naive_threshold_sig/mod.rs
```diff
@@ -267,56 +291,57 @@ pub fn generate_parameters(max_pks: usize) -> Result<Parameters<MNT4>, Synthesis
...
 
     fn generate_test_proof(
         max_pks:                  usize,
         valid_sigs:               usize,
         threshold:                usize,
         wrong_pks_threshold_hash: bool,
         wrong_wcert_sysdata_hash: bool,
-        params:                   Parameters<MNT4>,
-    ) -> Result<(Proof<MNT4>, Vec<MNT4Fr>), SynthesisError> {
+        params:                   (IndexProverKey<Fr, IPAPC>, IndexVerifierKey<Fr, IPAPC>),
```
DanieleDiBenedetto, line 326: You should need only the prover key for this function no ?

demo-circuit/src/naive_threshold_sig/mod.rs
```diff
@@ -402,29 +428,31 @@ mod test {
     }
 
     #[test]
+    #[should_panic]
```
DanieleDiBenedetto, line 431: I don't like this for a couple of reasons:
1) You don't execute all the test cases but the test will finish at the first panic (i.e. line 446)
2) You don't really know if the failure has happened for the expected reason or for some other that you don't know (i.e. a bug, that you would like to know).
So we have a couple of solutions here:
- You set debug_assertions = false, remove the #[should_panic] and assert the failure as it was before, but I don't like the debug assertions to be false (we may have other things that we are interested in asserting) and the problem of the #[should_panic] stays the same;
- We keep the debug assert to true and we make a separate test marked with #[should_panic] out of each of the test cases;
- The debug assertions that are causing the failures should be in https://github.com/HorizenLabs/marlin/blob/dev/src/ahp/mod.rs#L109. I think we can remove them at all.

I prefer the last solution honestly

jni/src/test/java/com/horizen/merkletreenative/MerkleTreeTest.java
```diff
@@ -1,314 +1,314 @@
...
+//        //Free memory
+//        zero.freeFieldElement();
+//        mht.freeInMemoryOptimizedMerkleTree();
+//        mhtCopy.freeInMemoryOptimizedMerkleTree();
+//        mhtRoot.freeFieldElement();
+//        mhtRootCopy.freeFieldElement();
+//    }
+//
+//    @Test
+//    public void testMerklePaths() {
```
DanieleDiBenedetto, line 142: Don't see any usage of hardcoded data here. Maybe we can keep it ?

jni/src/test/java/com/horizen/merkletreenative/MerkleTreeTest.java
```diff
@@ -1,314 +1,314 @@
...
+//        // Free memory
+//        mht.freeInMemoryOptimizedMerkleTree();
+//        mhtRoot.freeFieldElement();
+//        for (FieldElement leaf: testLeaves)
+//            leaf.freeFieldElement();
+//
+//    }
+//
+//    @Test
+//    public void testNonEmptyRightmost() {
```
DanieleDiBenedetto, line 212: Don't see any usage of hardcoded data here. Maybe we can keep it ?

jni/src/test/java/com/horizen/schnorrnative/SchnorrKeyPairTest.java
```diff
@@ -1,93 +1,93 @@
...
+//        )
+//        {
+//            assertNotNull("sk deserialization must not fail", sk);
+//            assertNotNull("message deserialization must not fail", message);
+//            assertTrue("Signature must be verified", keyPair.getPublicKey().verifySignature(sig, message));
+//        }
+//
+//    }
+//
+//    @Test
```
DanieleDiBenedetto, line 68: Don't see any usage of hardcoded data here. Maybe we can keep it ?

jni/src/test/java/com/horizen/sigproofnative/NaiveThresholdSigProofTest.java
```diff
@@ -1,314 +1,314 @@
...
+//import static org.junit.Assert.assertTrue;
+//import static org.junit.Assert.assertFalse;
+//
+//public class NaiveThresholdSigProofTest {
+//
+//    static int keyCount = 3;
+//    static long threshold = 2;
+//    static int backwardTransferCout = 10;
+//
+//    byte[] endEpochBlockHash = new byte[16];
```
DanieleDiBenedetto, line 28: 32 bytes

jni/src/test/java/com/horizen/sigproofnative/NaiveThresholdSigProofTest.java
```diff
@@ -1,314 +1,314 @@
...
+//                SchnorrSignature sig = new SchnorrSignature();
+//                signatureList.add(sig);
+//            }
+//        }
+//
+//        createAndVerifyProof();
+//    }
+//
+//    @Test
+//    public void testCreateRandomProof() throws Exception {
```
DanieleDiBenedetto, line 227: Don't see any usage of hardcoded data here. Maybe we can keep it ?

jni/src/test/java/com/horizen/vrfnative/VRFKeyPairTest.java
```diff
@@ -1,127 +1,127 @@
...
+//import static org.junit.Assert.assertNull;
+//import static org.junit.Assert.assertNotNull;
+//import static org.junit.Assert.assertTrue;
+//import static org.junit.Assert.assertEquals;
+//
+//public class VRFKeyPairTest {
+//
+//
+//    @Test
+//    public void testGenerate() throws Exception {
```
DanieleDiBenedetto, line 17: Don't see any usage of hardcoded data here. Maybe we can keep it ?

jni/src/test/java/com/horizen/schnorrnative/SchnorrKeyPairTest.java
```diff
@@ -1,93 +1,93 @@
...
+//import org.junit.Test;
+//
+//import java.util.Arrays;
+//
+//import static org.junit.Assert.*;
+//
+//public class SchnorrKeyPairTest {
+//
+//    @Test
+//    public void testGenerate() throws Exception {
```
DanieleDiBenedetto, line 13: Don't see any usage of hardcoded data here. Maybe we can keep it ?

jni/src/test/java/com/horizen/vrfnative/VRFKeyPairTest.java
```diff
@@ -1,127 +1,127 @@
...
+//            {
+//                assertNotNull("VRF Proof verification and VRF Output computation has failed.", vrfOutput);
+//
+//                // Check vrfOutput == expectedVrfOutput
+//                assertEquals("vrfOutput and expectedVrfOutput must be equal", vrfOutput, expectedVrfOutput);
+//            }
+//        }
+//    }
+//
+//    @Test
```
DanieleDiBenedetto, line 98: Don't see any usage of hardcoded data here. Maybe we can keep it ?

DanieleDiBenedetto (changes requested):
In general ok, but there were some Java tests not depending on any hardcoded data that you should uncomment and test (don't forget the functions annotated with @Before and @After that are important and also used for tests that don't depend on hardcoded data.
If you were able to execute the test you would've been able to notice some UnsatisfiedLink exception: I assume you took the code from deserialization_checked_unchecked branch but you imported those changes only Rust side there also lot of changes in Java classes and Java unit test: you should import them too

jni/src/test/java/com/horizen/sigproofnative/NaiveThresholdSigProofTest.java
phoinic: With epoch 32 I am getting the error in UT: "Attempt to deserialize a field element over the modulus"

jni/src/test/java/com/horizen/merkletreenative/MerkleTreeTest.java
phoinic: This test uses expectedRootBytes hardcoded array

jni/src/test/java/com/horizen/merkletreenative/MerkleTreeTest.java
phoinic: This test uses expectedRootBytes hardcoded array

jni/src/test/java/com/horizen/schnorrnative/SchnorrKeyPairTest.java
phoinic: Yes, this test is working, I've uncommented it

jni/src/test/java/com/horizen/vrfnative/VRFKeyPairTest.java
phoinic: Yes, this test is working, I've uncommented it

jni/src/test/java/com/horizen/schnorrnative/SchnorrKeyPairTest.java
phoinic: Yes, this test is working, I've uncommented it

jni/src/test/java/com/horizen/vrfnative/VRFKeyPairTest.java
phoinic: Yes, this test is working, I've uncommented it

api/src/ginger_calls.rs
phoinic: With epoch 32 bytes length I am getting the error in UT: "Attempt to deserialize a field element over the modulus". Actually tweedle fields are 256bit long and with 256bit vector we of course can have value which is bigger then modulus of the field, The FromBytes implementation for fields in ginger-lib cannot do reduce by modulus during the reading from buffer. What should be the solution?

1. Change the implementation in ginger-lib so if modulus smaller than passed bytes, just do the reduction

2. Same thing as above, but do it in the zendoo in function "read_field_element_from_buffer_with_padding"

I prefer second variant, because first one will affect all projects based on ginger-lib in unpredictable way

DanieleDiBenedetto: Yes you're completely right. There can be several approaches: 

1.  We truncate the byte array by 1 byte (or 2 bit) so we are able to safely deserialize a FieldElement (we lose just 4 bit/1 bit of security);
2.  We call the call the to_field_elements() function of GingerLib for [u8] (https://github.com/HorizenOfficial/ginger-lib/blob/algebra_ocl/algebra/src/to_field_vec.rs#L161), but this will deserialize 2 field elements out of this 32 byte array (and will slightly change the circuit too);
3.  Deserialize two field elements with the aforementioned function, compute the PoseidonHash and use the output as "deserialized" field element.

We need to discuss with MC and SC what is the preferred approach. As I understood they have not still reached the point in which this MC hashes are used, but in other points they need to deserialize a FieldElement from a SHA-256 hash and they are using approach number 1. 
Let's discuss it separately.

jni/src/test/java/com/horizen/sigproofnative/NaiveThresholdSigProofTest.java
phoinic: Done

api/src/ginger_calls.rs
```diff
@@ -200,6 +200,9 @@ pub fn read_field_element_from_buffer_with_padding(buffer: &[u8]) -> IoResult<Fi
     new_buffer.extend_from_slice(buffer);
     for _ in buff_len..FIELD_SIZE { new_buffer.push(0u8) } //Add padding zeros to reach field size
 
+    //Truncate buffer to fit the MODULUS
```
DanieleDiBenedetto, line 203: I wouldn't put this in the code Rust side: it's the caller Java side who should be responsible of passing a correct byte array

DanieleDiBenedetto (changes requested):

api/src/ginger_calls.rs
phoinic: Done

api/src/ginger_calls.rs
phoinic: In ginger-lib we have PHANTOM_MERKLE_ROOT defined only for MNT4, what the correct parameter for it for tweedle?

api/src/ginger_calls.rs
```diff
@@ -1,85 +1,111 @@
...
-        batched_crh::MNT4BatchPoseidonHash as BatchFieldHash,
-    },
+    poseidon::parameters::tweedle::{TweedleFrPoseidonHash as PoseidonHash, TweedleFrBatchPoseidonHash as BatchFieldHash},
     FieldBasedHash,
     bowe_hopwood::{
         BoweHopwoodPedersenCRH, BoweHopwoodPedersenParameters
     },
 }, merkle_tree::field_based_mht::{
-    smt::{BigMerkleTree, LazyBigMerkleTree, Coord, OperationLeaf},
+    // smt::{BigMerkleTree, LazyBigMerkleTree, Coord, OperationLeaf},
```
i-Alex, line 20: @DanieleDiBenedetto so now we have commented BigMerkleTree and no dependency on `smt` feature of ginger lib.
Am I right, that now there is no dependency on rocks DB, so no problem with libstd dependencies on Linux and Windows?

api/src/lib.rs
```diff
@@ -493,6 +515,22 @@ pub extern "system" fn Java_com_horizen_schnorrnative_SchnorrSignature_nativeDes
     *sig_object
 }
 
+#[no_mangle]
+pub extern "system" fn Java_com_horizen_schnorrnative_SchnorrSignature_nativeIsValidSignature(
```
i-Alex, line 519: is not used in Java `SchnorrSignature` class.

api/src/lib.rs
i-Alex: Agree, let's comment all `com.horizen.merkletreenative` package classes related JNIs.

i-Alex (changes requested):

api/src/ginger_calls.rs
```diff
@@ -1,85 +1,117 @@
...
+// }
+
+type IPAPC = InnerProductArgPC<PCAffine, Blake2s>;
+// type MarlinInst = Marlin<Fr, IPAPC, Blake2s, MarlinNoLCNoZk>;
+type MarlinInst = Marlin<Fr, IPAPC, Blake2s>;
+
+pub const FIELD_SIZE: usize = 32; //Field size in bytes
+pub const SCALAR_FIELD_SIZE: usize = FIELD_SIZE;// 32
 pub const G1_SIZE: usize = 193;
 pub const G2_SIZE: usize = 385;
```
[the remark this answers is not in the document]
phoinic: G1_SIZE and G2_SIZE constants removed, and G_SIZE constant introduced instead

api/src/ginger_calls.rs
phoinic: Done

jni/src/test/java/com/horizen/merkletreenative/MerkleTreeTest.java
```diff
@@ -45,265 +45,264 @@ public void initTestParams() {
...
-        //Free memory
-        lazySmt.freeAndDestroyLazyMerkleTree();
-        lazySmtRoot.freeFieldElement();
-
-        // Check data have been destroyed
-        f = new File("./db_big_lazy_persistency");
-        assertTrue("DB has not been destroyed", !f.exists());
-    }
+//    @Test
+//    public void testMerkleTrees() {
```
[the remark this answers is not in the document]
phoinic: Done

jni/src/test/java/com/horizen/poseidonnative/PoseidonHashTest.java
```diff
@@ -1,78 +1,78 @@
...
-}
+//package com.horizen.poseidonnative;
+//
+//import com.horizen.librustsidechains.FieldElement;
+//import org.junit.Test;
+//
+//import static org.junit.Assert.*;
+//
+//
+//public class PoseidonHashTest {
```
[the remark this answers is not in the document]
phoinic: Done

api/src/lib.rs
phoinic: This function is commented

jni/src/test/java/com/horizen/vrfnative/VRFKeyPairTest.java
```diff
@@ -24,76 +24,76 @@ public void testGenerate() throws Exception {
...
-            {
-                assertNotNull("VRF Proof verification and VRF Output computation has failed.", vrfOutput);
-
-                // Check vrfOutput == expectedVrfOutput
-                assertEquals("vrfOutput and expectedVrfOutput must be equal", vrfOutput, expectedVrfOutput);
-            }
-        }
-    }
+//    @Test
+//    public void testProveVerify() throws Exception {
```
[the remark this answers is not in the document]
phoinic: Where can I take them ?

jni/src/test/java/com/horizen/sigproofnative/NaiveThresholdSigProofTest.java
phoinic: This test depends on "samples" stored in jni/src/test/resources. Can I regenerate them somehow ?

api/src/ginger_calls.rs
DanieleDiBenedetto: I will recompute it in future, stick with this now

api/src/ginger_calls.rs
DanieleDiBenedetto: A quick note on this, @i-Alex please I would like your feedback too.
The size of a Marlin proof may change according to the specific circuit for which the proof is created and according to some parameters that should be declared at SC setup time.
My proposal is that, to know the serialized size of a proof, we can expose an additional function get_sc_proof_size() to be called before serializing, and SDK size this value can be saved somewhere (it won't change for the single SC, the problem is that it may change from SC to SC).

jni/src/test/java/com/horizen/sigproofnative/NaiveThresholdSigProofTest.java
DanieleDiBenedetto: Yes, the test at https://github.com/HorizenOfficial/zendoo-sc-cryptolib/blob/updatable_poseidon_tweedle/api/src/ginger_calls.rs#L680 should generate and write to file the pk/vk/proof that I remember can be used for this test

api/Cargo.toml
```diff
@@ -7,9 +7,11 @@ edition = "2018"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-algebra = {version = "0.1.0", git = "https://github.com/ZencashOfficial/ginger-lib.git", branch = "smt"}
-primitives = {version = "0.1.0", features = ["smt", "signature", "vrf"], git = "https://github.com/ZencashOfficial/ginger-lib.git", branch = "smt"}
-proof-systems = {version = "0.1.0", features = ["groth16"], git = "https://github.com/ZencashOfficial/ginger-lib.git", branch = "smt"}
+algebra = {version = "0.1.0", features = ["mnt4_753", "mnt6_753"], git = "https://github.com/HorizenOfficial/ginger-lib.git", branch = "algebra_ocl_serializations"}
```
DanieleDiBenedetto, line 10: Remove the MNTs from the features

api/Cargo.toml
```diff
@@ -7,9 +7,11 @@ edition = "2018"
...
+primitives = {version = "0.1.0", features = ["mnt4_753", "mnt6_753", "merkle_tree", "smt"], git = "https://github.com/HorizenOfficial/ginger-lib.git", branch = "algebra_ocl_serializations"}
+proof-systems = {version = "0.1.0", features = ["groth16"], git = "https://github.com/HorizenOfficial/ginger-lib.git", branch = "algebra_ocl_serializations"}
```
DanieleDiBenedetto, line 12: This can be removed totally, we don't use groth16 anymore

demo-circuit/Cargo.toml
```diff
@@ -6,19 +6,22 @@ edition = "2018"
...
-r1cs-core = {version = "0.1.0", git = "https://github.com/ZencashOfficial/ginger-lib.git", branch = "smt"}
-r1cs-std = {version = "0.1.0", git = "https://github.com/ZencashOfficial/ginger-lib.git", branch = "smt"}
-proof-systems = {version = "0.1.0", features = ["groth16"], git = "https://github.com/ZencashOfficial/ginger-lib.git", branch = "smt"}
+algebra = {version = "0.1.0", features = ["mnt4_753", "mnt6_753", "tweedle"], git = "https://github.com/HorizenOfficial/ginger-lib.git", branch = "algebra_ocl_serializations"}
+primitives = {version = "0.1.0", features = ["mnt4_753", "mnt6_753", "tweedle", "signature"], git = "https://github.com/HorizenOfficial/ginger-lib.git", branch = "algebra_ocl_serializations"}
+proof-systems = {version = "0.1.0", features = ["groth16"], git = "https://github.com/HorizenOfficial/ginger-lib.git", branch = "algebra_ocl_serializations"}
+r1cs-crypto = {version = "0.1.0", features = ["mnt4_753", "mnt6_753", "tweedle", "signature"], git = "https://github.com/HorizenOfficial/ginger-lib.git", branch = "algebra_ocl_serializations"}
```
DanieleDiBenedetto, line 12: Remove MNTs

demo-circuit/Cargo.toml
```diff
@@ -6,19 +6,22 @@ edition = "2018"
...
+r1cs-core = {version = "0.1.0", git = "https://github.com/HorizenOfficial/ginger-lib.git", branch = "algebra_ocl_serializations"}
+r1cs-std = {version = "0.1.0", features = ["mnt4_753", "mnt6_753", "tweedle"], git = "https://github.com/HorizenOfficial/ginger-lib.git", branch = "algebra_ocl_serializations"}
```
DanieleDiBenedetto, line 14: Remove MNTs

ouroboros/Cargo.toml
```diff
@@ -5,7 +5,6 @@ authors = ["DanieleDiBenedetto <daniele@horizenlabs.io>"]
 edition = "2018"
 
 [dependencies]
-algebra = {version = "0.1.0", git = "https://github.com/ZencashOfficial/ginger-lib.git", branch = "smt"}
-primitives = {version = "0.1.0", features = ["vrf"], git = "https://github.com/ZencashOfficial/ginger-lib.git", branch = "smt"}
+algebra = {version = "0.1.0", features = ["mnt4_753", "mnt6_753", "tweedle"], git = "https://github.com/HorizenOfficial/ginger-lib.git", branch = "algebra_ocl_serializations"}
```
DanieleDiBenedetto, line 8: Remove MNTs

ouroboros/Cargo.toml
```diff
@@ -5,7 +5,6 @@ authors = ["DanieleDiBenedetto <daniele@horizenlabs.io>"]
...
+primitives = {version = "0.1.0", features = ["mnt4_753", "mnt6_753", "tweedle", "vrf"], git = "https://github.com/HorizenOfficial/ginger-lib.git", branch = "algebra_ocl_serializations"}
```
DanieleDiBenedetto, line 9: Remove MNTs

DanieleDiBenedetto (changes requested):

ouroboros/Cargo.toml
phoinic: Done

ouroboros/Cargo.toml
phoinic: Done

demo-circuit/Cargo.toml
phoinic: Done

demo-circuit/Cargo.toml
phoinic: Done

demo-circuit/Cargo.toml
```diff
@@ -6,19 +6,22 @@ edition = "2018"
...
-primitives = {version = "0.1.0", features = ["signature"], git = "https://github.com/ZencashOfficial/ginger-lib.git", branch = "smt"}
-r1cs-crypto = {version = "0.1.0", features = ["signature"], git = "https://github.com/ZencashOfficial/ginger-lib.git", branch = "smt"}
-r1cs-core = {version = "0.1.0", git = "https://github.com/ZencashOfficial/ginger-lib.git", branch = "smt"}
-r1cs-std = {version = "0.1.0", git = "https://github.com/ZencashOfficial/ginger-lib.git", branch = "smt"}
-proof-systems = {version = "0.1.0", features = ["groth16"], git = "https://github.com/ZencashOfficial/ginger-lib.git", branch = "smt"}
+algebra = {version = "0.1.0", features = ["mnt4_753", "mnt6_753", "tweedle"], git = "https://github.com/HorizenOfficial/ginger-lib.git", branch = "algebra_ocl_serializations"}
+primitives = {version = "0.1.0", features = ["mnt4_753", "mnt6_753", "tweedle", "signature"], git = "https://github.com/HorizenOfficial/ginger-lib.git", branch = "algebra_ocl_serializations"}
+proof-systems = {version = "0.1.0", features = ["groth16"], git = "https://github.com/HorizenOfficial/ginger-lib.git", branch = "algebra_ocl_serializations"}
```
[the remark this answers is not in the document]
phoinic: Done

demo-circuit/Cargo.toml
```diff
@@ -6,19 +6,22 @@ edition = "2018"
...
 
 [dependencies]
-algebra = {version = "0.1.0", git = "https://github.com/ZencashOfficial/ginger-lib.git", branch = "smt"}
-primitives = {version = "0.1.0", features = ["signature"], git = "https://github.com/ZencashOfficial/ginger-lib.git", branch = "smt"}
-r1cs-crypto = {version = "0.1.0", features = ["signature"], git = "https://github.com/ZencashOfficial/ginger-lib.git", branch = "smt"}
-r1cs-core = {version = "0.1.0", git = "https://github.com/ZencashOfficial/ginger-lib.git", branch = "smt"}
-r1cs-std = {version = "0.1.0", git = "https://github.com/ZencashOfficial/ginger-lib.git", branch = "smt"}
-proof-systems = {version = "0.1.0", features = ["groth16"], git = "https://github.com/ZencashOfficial/ginger-lib.git", branch = "smt"}
+algebra = {version = "0.1.0", features = ["mnt4_753", "mnt6_753", "tweedle"], git = "https://github.com/HorizenOfficial/ginger-lib.git", branch = "algebra_ocl_serializations"}
+primitives = {version = "0.1.0", features = ["mnt4_753", "mnt6_753", "tweedle", "signature"], git = "https://github.com/HorizenOfficial/ginger-lib.git", branch = "algebra_ocl_serializations"}
```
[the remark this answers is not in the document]
phoinic: Done

demo-circuit/Cargo.toml
```diff
@@ -6,19 +6,22 @@ edition = "2018"
 include = ["Cargo.toml", "src"]
 
 [dependencies]
-algebra = {version = "0.1.0", git = "https://github.com/ZencashOfficial/ginger-lib.git", branch = "smt"}
-primitives = {version = "0.1.0", features = ["signature"], git = "https://github.com/ZencashOfficial/ginger-lib.git", branch = "smt"}
-r1cs-crypto = {version = "0.1.0", features = ["signature"], git = "https://github.com/ZencashOfficial/ginger-lib.git", branch = "smt"}
-r1cs-core = {version = "0.1.0", git = "https://github.com/ZencashOfficial/ginger-lib.git", branch = "smt"}
-r1cs-std = {version = "0.1.0", git = "https://github.com/ZencashOfficial/ginger-lib.git", branch = "smt"}
-proof-systems = {version = "0.1.0", features = ["groth16"], git = "https://github.com/ZencashOfficial/ginger-lib.git", branch = "smt"}
+algebra = {version = "0.1.0", features = ["mnt4_753", "mnt6_753", "tweedle"], git = "https://github.com/HorizenOfficial/ginger-lib.git", branch = "algebra_ocl_serializations"}
```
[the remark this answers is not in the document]
phoinic: Done

api/Cargo.toml
phoinic: Done

api/Cargo.toml
```diff
@@ -7,9 +7,11 @@ edition = "2018"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-algebra = {version = "0.1.0", git = "https://github.com/ZencashOfficial/ginger-lib.git", branch = "smt"}
-primitives = {version = "0.1.0", features = ["smt", "signature", "vrf"], git = "https://github.com/ZencashOfficial/ginger-lib.git", branch = "smt"}
-proof-systems = {version = "0.1.0", features = ["groth16"], git = "https://github.com/ZencashOfficial/ginger-lib.git", branch = "smt"}
+algebra = {version = "0.1.0", features = ["mnt4_753", "mnt6_753"], git = "https://github.com/HorizenOfficial/ginger-lib.git", branch = "algebra_ocl_serializations"}
+primitives = {version = "0.1.0", features = ["mnt4_753", "mnt6_753", "merkle_tree", "smt"], git = "https://github.com/HorizenOfficial/ginger-lib.git", branch = "algebra_ocl_serializations"}
```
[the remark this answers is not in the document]
phoinic: Done

api/Cargo.toml
phoinic: Done

jni/src/test/java/com/horizen/sigproofnative/NaiveThresholdSigProofTest.java
phoinic: Ok, thank you, but sample_pk now has 1.2 GB size, and sample_vk 130 MB. Is it Ok to keep them in GIT?

DanieleDiBenedetto: I pushed a fix in algebra_ocl_serializations yesterday, now the sizes should be much smaller. I suggest (temporarily) to change the code of function generate_params (see related comment)

demo-circuit/src/naive_threshold_sig/mod.rs
```diff
@@ -267,56 +291,57 @@ pub fn generate_parameters(max_pks: usize) -> Result<Parameters<MNT4>, Synthesis
         _field:                   PhantomData
     };
 
-    let params = generate_random_parameters::<MNT4, _, _>(c, &mut rng);
-    params
+    let universal_srs = MarlinInst::universal_setup(33254, 33254, 812307, &mut rng).unwrap();
```
DanieleDiBenedetto, line 294:
```suggestion
    // Temporary workaround to get the sizes, in future we will only call the index function
    // using the DLOG keys generated at bootstrap
    let info = marlin::AHPForR1CS::<Fr>::index(c.clone()).unwrap().index_info;

    let universal_srs = MarlinInst::universal_setup(
        info.num_constraints,
        info.num_variables,
        info.num_non_zero,
        &mut rng
    ).unwrap();
```

DanieleDiBenedetto (changes requested):

jni/src/test/java/com/horizen/vrfnative/VRFKeyPairTest.java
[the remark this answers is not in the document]
DanieleDiBenedetto: Let's restore it later, as we changed the hash implementation in Ginger as a result of a code review, so this will be changed again in the future

jni/src/test/java/com/horizen/sigproofnative/NaiveThresholdSigProofTest.java
DanieleDiBenedetto: This will allow to upload them to git

api/src/ginger_calls.rs
DanieleDiBenedetto: (If needed)

demo-circuit/src/naive_threshold_sig/mod.rs
phoinic: This part is added

jni/src/test/java/com/horizen/vrfnative/VRFKeyPairTest.java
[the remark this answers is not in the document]
phoinic: So for now we are keeping it commented, right?

api/src/lib.rs
phoinic: Done

api/src/lib.rs
phoinic: Done

api/src/lib.rs
phoinic: Done

jni/src/test/java/com/horizen/sigproofnative/NaiveThresholdSigProofTest.java
phoinic: Hm, In GIT there is a limit for max file size: 100 MB. And current PK is 346 MB =(

demo-circuit/src/naive_threshold_sig/mod.rs
phoinic: Done

jni/src/test/java/com/horizen/vrfnative/VRFKeyPairTest.java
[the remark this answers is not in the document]
DanieleDiBenedetto: Yes

jni/src/test/java/com/horizen/sigproofnative/NaiveThresholdSigProofTest.java
DanieleDiBenedetto: Let's generate them directly Java side, execute the test and then delete them.

api/src/ginger_calls.rs
phoinic: Variable "ZK_PROOF_SIZE" removed, and size of proof now is not fixed

jni/src/test/java/com/horizen/sigproofnative/NaiveThresholdSigProofTest.java
phoinic: Done

demo-circuit/src/naive_threshold_sig/mod.rs
phoinic: Ok, I have implemented the last solution inside the marlin in branch "dev_serializations". "shoud_panic" removed from UTs

api/src/lib.rs
```diff
@@ -2431,6 +2433,37 @@ pub extern "system" fn Java_com_horizen_sigproofnative_NaiveThresholdSigProof_na
...
+fn write_to_file<T: ToBytes>(to_write: &T, file_path: &str) -> IoResult<()>{
+    let mut fs = File::create(file_path)?;
+    to_write.write(&mut fs)?;
+    Ok(())
+}
+
+#[no_mangle]
+pub extern "system" fn Java_com_horizen_sigproofnative_NaiveThresholdSigProof_nativeSetup(
+    _env: JNIEnv,
+    _params: JObject,
```
DanieleDiBenedetto, line 2445: Does it work with this _params as JObject ? It's a static function so it should be a JClass

DanieleDiBenedetto (changes requested):

api/src/lib.rs
phoinic: Done

DanieleDiBenedetto (approved):

i-Alex (approved):
Keep in mind there are some Java regression tests to be restored.

albertog78 (changes requested):

albertog78 (approved):

Issue #38, opened 2020-08-06
Fix Java deserialization checks


DanieleDiBenedetto:
If deserialization of a Rust struct fails, Rust side, the corresponding Java object is created with a pointer to null, instead of returning a whole null Java Object.
The assertions Java side assume the second behaviour instead.

DanieleDiBenedetto:
Done as part of #46

PR #52, opened 2021-06-08
Bitvector and deserialization interfaces updates.

i-Alex:


api/src/lib.rs
```diff
@@ -3094,4 +3094,34 @@ pub extern "system" fn Java_com_horizen_librustsidechains_Utils_nativeCompressed
...
+
+    // Compute merkle_root
+    match merkle_root_from_compressed_bytes(compressed_bit_vector.as_slice(), expected_uncompressed_size) {
+        Ok(merkle_root) => {
+            // Return merkle_root bytes
+            let merkle_root_bytes = serialize_to_buffer(&merkle_root).expect("Should be able to serialize merkle_root");
+            _env.byte_array_from_slice(merkle_root_bytes.as_slice()).expect("Cannot write jobject.")
+        }
+        Err(_) => {
+            _env.throw_new("java/lang/Exception", "Cannot compute merkle root with size check.").expect("Exception expected.");
```
DanieleDiBenedetto, line 3121: At a certain point we need to address issue #38

jni/src/main/java/com/horizen/vrfnative/VRFPublicKey.java
```diff
@@ -32,6 +32,10 @@ public static VRFPublicKey deserialize(byte[] publicKeyBytes, boolean checkPubli
     return nativeDeserializePublicKey(publicKeyBytes, checkPublicKey);
   }
 
+  public static VRFPublicKey deserialize(byte[] publicKeyBytes) {
```
DanieleDiBenedetto, line 35: You need this default "deserialize" only for SchnorrPk, SchnorrSignature and VrfPk ? Nothing else ? Like SNARK pk, vk, proof, or Vrf proof and so on

DanieleDiBenedetto (approved):

jni/src/main/java/com/horizen/vrfnative/VRFPublicKey.java
i-Alex: I forgot about vrf proof. Will add it.
About snark proofs and keys it is fine to keep as is.

albertog78 (approved):

PR #53, opened 2021-06-14
More options when serializing/deserializing

DanieleDiBenedetto:


api/src/lib.rs
```diff
@@ -47,8 +38,9 @@ fn read_nullable_raw_pointer<'a, T>(input: *const T) -> Option<&'a T> {
 
 fn serialize_from_raw_pointer<T: CanonicalSerialize>(
     to_write: *const T,
+    compressed: Option<bool>,
```
i-Alex, line 41: honestly, boolean option seems very strange. I can understand the meaning of true/false, but what is the meaning of `None`? Logically `None` is the same as `Some(false)`.
The same in other places with `compressed` arg.

i-Alex (changes requested):

api/src/lib.rs
DanieleDiBenedetto: You're right, it's more a conceptual meaning. Some types (like FieldElement) can't be compressed or doesn't have any semantic check: I found for these a little bit ugly to pass a boolean value that doesn't have any effect or doesn't have any meaning. I thought passing None, instead, would've been more self-explanatory

DanieleDiBenedetto: If you look at the internal implementation in cctp-lib, in fact, it will be unwrapped to false if the Option is None, but at least this logic is hidden from the user and who reads the code knows that no check/compression is involved for a given type (or at least this was may aim)

i-Alex: If FieldElement has no compressed form or no checks, it means that the serialization interface for it should be different to the current one. Consider the case when for the FE serialization someone will set `compressed = Some(true)`, which is allowed by the interface. The problem still exists.
I would suggest to review the interfaces for the serialization.

i-Alex (approved):
Approved. Decided to check interfaces later.

albertog78 (approved):

PR #57, opened 2021-06-21
Check on proof and vk size

DanieleDiBenedetto:
Setup function has been modified to take additional parameters needed to check whether the circuit for which the user is about to create the proving key, verification key and for which it will create a proof, will respect a maximum size declared elsewhere (in our case, the MC will declare it)


i-Alex (approved):
SDK notes: during sc-cryptolib integration check for the values of `max_proof_size` and `max_vk_size` to be consistent with the MC.

PR #59, opened 2021-07-07
Restored commented Java tests

DanieleDiBenedetto:
Removed hardcoded proof tests too, they didn't make sense as now we generate pk, vk and proof each time from scratch.
@i-Alex do you see other tests that might make sense to add here ?


i-Alex (approved):
Seems that all tests were restored. As for me, current test suit is enough and covers the logic on proper level.
For the future, when we will modify the rust jni part to avoid do `panic` on failures, we may extend the tests suit with the negative cases.

PR #61, opened 2021-08-11
Wrapping up possible crashes 

phoinic:
Wrapped cases:

- "unwrap" usage - can cause panics
- assertions in parameters check - can cause panics
- direct calls of "panic!"

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ incremental = true
 panic = "unwind"
 
 # Uncomment these lines for local development paths
-
+#
 #[patch.'https://github.com/HorizenOfficial/ginger-lib.git']
 #algebra = { path = '../ginger-lib/algebra' }
 #primitives = { path = '../ginger-lib/primitives' }
@@ -33,3 +33,9 @@ panic = "unwind"
 #
 #[patch.'https://github.com/HorizenOfficial/zendoo-cctp-lib.git']
 #cctp_primitives = { path = '../zendoo-cctp-lib/cctp_primitives' }
+#
+#[patch.'https://github.com/HorizenLabs/marlin']
+#marlin = { path = '../marlin' }
+#
+#[patch.'https://github.com/HorizenLabs/poly-commit']
+#poly-commit = { path = '../poly-commit' }
--- a/api/Cargo.toml
+++ b/api/Cargo.toml
@@ -7,10 +7,10 @@ edition = "2018"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-algebra = {version = "0.1.0", features = ["tweedle"], git = "https://github.com/HorizenOfficial/ginger-lib.git", branch = "audit_fixes"}
-primitives = {version = "0.1.0", features = ["tweedle", "merkle_tree"], git = "https://github.com/HorizenOfficial/ginger-lib.git", branch = "audit_fixes"}
+algebra = {version = "0.1.0", features = ["tweedle"], git = "https://github.com/HorizenOfficial/ginger-lib.git", branch = "audit_fixes_005"}
+primitives = {version = "0.1.0", features = ["tweedle", "merkle_tree"], git = "https://github.com/HorizenOfficial/ginger-lib.git", branch = "audit_fixes_005"}
 
-cctp_primitives = { git = "https://github.com/HorizenOfficial/zendoo-cctp-lib.git", branch = "audit_fixes" }
+cctp_primitives = { git = "https://github.com/HorizenOfficial/zendoo-cctp-lib.git", branch = "audit_fixes_005" }
 
 #ouroboros = {path = "../ouroboros"}
 demo-circuit = {path = "../demo-circuit"}
--- a/api/src/cctp_calls.rs
+++ b/api/src/cctp_calls.rs
@@ -199,6 +199,8 @@ pub fn create_naive_threshold_sig_proof(
         Some(compressed_pk)
     )?;
 
+    let g1_ck = get_g1_committer_key()?;
+
     let proof = match pk {
         ZendooProverKey::Darlin(_) => unimplemented!(),
         ZendooProverKey::CoboundaryMarlin(pk) => {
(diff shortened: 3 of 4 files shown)


DanieleDiBenedetto (changes requested):
We have many calls to expect() that is like an unwrap. So:

- Either we wrap them too;
- We leave them as they are and we use the catch_unwind strategy that we should do also in zendoo-mc-cryptolib (that I'm afraid we need to do here anyway).

What do you think @phoinic ?

DanieleDiBenedetto:
@i-Alex if you can take a look at it too would be great. Some interfaces have slightly changed (mostly returning Err or Option because we removed some dangerous unwraps).
A parallel PR is opened on zendoo-sc-cryptolib too (https://github.com/HorizenOfficial/zendoo-sc-cryptolib/pull/61).
Please be aware of these changes when reporting them into the SDK (shouldn't be a huge deal).

DanieleDiBenedetto (approved):

api/src/lib.rs
```diff
@@ -1180,17 +1182,17 @@ pub extern "system" fn Java_com_horizen_merkletreenative_InMemoryOptimizedMerkle
         read_raw_pointer(t.j().unwrap() as *const GingerMHT)
     };
 
-    let tree_copy = finalize_ginger_mht(tree);
-
-    return_jobject(&_env, tree_copy, "com/horizen/merkletreenative/InMemoryOptimizedMerkleTree")
-        .into_inner()
+    match finalize_ginger_mht(tree) {
+        Ok(tree_copy) => return_jobject(&_env, tree_copy, "com/horizen/merkletreenative/InMemoryOptimizedMerkleTree").into_inner(),
+        Err(_) => return std::ptr::null::<jobject>() as jobject //CRYPTO_ERROR
```
i-Alex, line 1187: for now is fine. Later when have time, we need to throw a java exception instead of returning the `null`. Also annotate the corresponding java method with `throws ...`
The issue is already present in a backlog.

i-Alex (approved):

api/src/lib.rs
```diff
@@ -39,9 +39,8 @@ fn read_nullable_raw_pointer<'a, T>(input: *const T) -> Option<&'a T> {
 fn serialize_from_raw_pointer<T: CanonicalSerialize>(
     to_write: *const T,
     compressed: Option<bool>,
-) -> Vec<u8> {
+) -> Result<Vec<u8>, SerializationError> {
     serialize_to_buffer(read_raw_pointer(to_write), compressed)
-        .expect(format!("unable to write {} to buffer", type_name::<T>()).as_str())
```
DanieleDiBenedetto, line 44: That's a pretty big change actually affecting also the SDK I assume. This function is used by "serialize_from_jobject" function that is used in a lot of places. So the caller should check each time that the return value is not NULL. 
Cleanest way to handle this is through the usage of Java exceptions but we would still need to add lot of throws to each function argument.
Since we don't have much time now, my suggestion would be to leave the .expect() here and put a TODO to handle this more carefully.
@i-Alex what do you think ?

api/src/lib.rs
```diff
@@ -2250,14 +2257,15 @@ pub extern "system" fn Java_com_horizen_commitmenttree_CommitmentTree_nativeAddS
...
+    // let constant_fe;
     let constant = if _constant_nullable.is_null() {
         Option::None
     } else {
         let constant_bytes = parse_jbyte_array_to_vec(&_env, &_constant_nullable, FIELD_SIZE);
-        constant_fe = FieldElement::deserialize(constant_bytes.as_slice()).expect("Can't parse the input constant_bytes into FieldElement");
-        Some(&constant_fe)
+        match FieldElement::deserialize(constant_bytes.as_slice()) {
+            Ok(constant_fe) => Option::Some(constant_fe),
+            Err(_) => Option::None
```
DanieleDiBenedetto, line 2267: This is not correct, this is a failure and we shouldn't let execution continue. Return false.

DanieleDiBenedetto (changes requested):
I would like to hear Oleksandr opinion.
In updating Java documentation and tests it seems you forgot to address the changes to the GingerMHT

api/src/lib.rs
i-Alex: I agree with Daniele to leave it as it was before.
There are a lot of places where we need to update the code (JNI part) to cover the "null cases".

phoinic: Ok, I've returned back the previous state and put TODO regarding Java exceptions for future

api/src/lib.rs
phoinic: Done

DanieleDiBenedetto (changes requested):
You need to undo too the comment in Java classes and tests you added in the previous commit.
Moreover, you still need to add the comments in the Java classes and tests related to the modifications you did to the Merkle Tree functions

DanieleDiBenedetto (approved):

i-Alex (approved):

DanieleDiBenedetto:
So I re-requested a review, let's give a little bit of context.
Each JNI Rust function is wrapped in a macro now, that:
- Add the FFI related metadata;
- Wrap the function body in a catch_unwind closure: if the code panics, a Java RuntimeException is automatically thrown (Rust won't abort);
- A throw() function was added, with the same behaviour of the Java one: the execution of the function body will be interrupted (using a panic and exploiting the fact that we are in a catch_unwind closure) and the control return to the Java side with the thrown exception (without aborting Rust side). The reason for this is that, normally, the exception is thrown Java side only when the execution of the native Rust function finishes;
-  If multiple exceptions are thrown from Rust side, only the first thrown exception is returned

This partially addresses #31 but must be completed with proper thrown of user-crafted exceptions in case of errors, instead of just printing and returning default values

jni/src/main/java/com/horizen/librustsidechains/Library.java
```diff
@@ -123,4 +123,11 @@ public static synchronized void load() {
         }
     }
 
+    // Used for testing purposes
+    private static native void nativePanickingFunction();
+
+    public static void panickingFunction() {
```
i-Alex, line 129: I believe we use it only for tests, so can be `protected static`

i-Alex (approved):

PR #70, opened 2021-11-17
InMemoryLazySparseMerkleTree

DanieleDiBenedetto:
Java classes and JNI functions Rust implementation


jni/src/main/java/com/horizen/merkletreenative/InMemoryLazySparseMerkleTree.java
```diff
@@ -0,0 +1,84 @@
...
+    public boolean isPositionEmpty(long position){
+        if (merkleTreePointer == 0)
+            throw new IllegalStateException("InMemoryLazySparseMerkleTree instance was freed.");
+        return nativeIsPositionEmpty(position);
+    }
+
+    private native FieldElement nativeAddLeaves(PositionLeaf[] leaves);
+
+    // Add leaves to tree, compute and return the root
+    public FieldElement addLeaves(List<PositionLeaf> leaves){
```
i-Alex, line 39: Why we should return root?
I would expect to add leaves and return boolean if it was successful

api/src/lib.rs
```diff
@@ -3318,4 +3349,218 @@ ffi_export!(
...
+        Err(e) => throw!(&_env, "java/lang/Exception", format!("Cannot insert leaves: {}", e.to_string()).as_str(), JObject::null().into_inner())
+    }    
+});
+
+ffi_export!(
+    fn Java_com_horizen_merkletreenative_InMemoryLazySparseMerkleTree_nativeRemoveLeaves(
+    _env: JNIEnv,
+    _tree: JObject,
+    _positions: jlongArray,
+) -> jobject
```
i-Alex, line 3476: Java part expect `void` as a result.
I think, we don't need to calculate the root as a last step. Only remove leafs if possible.

i-Alex (changes requested):

i-Alex (approved):
Perfect!

Yousefatm999:
Good

Yousefatm999 (approved):

albertog78 (approved):

i-Alex (approved):

albertog78 (approved):

Issue #72, opened 2021-11-24
Update circuit setup interfaces to new PROOF_PLUS_VK_SIZE threshold


DanieleDiBenedetto:
Currently, the setup() function in NaiveThresholdSigProof class accepts maxProofSize and maxVkSize; In reality those don't really matter, what matters is their sum (MC will check proof+vk size being below a certain hardcoded threshold).
Update these functions (also in the SDK) accordingly

DanieleDiBenedetto:
Fixed as part of #73

PR #73, opened 2021-11-24
CSW 

DanieleDiBenedetto:
- Introduced CSW circuit and all the Java classes and Rust/Java functions to perform setup/creation/verification;
- Fix issue #72 ;
- Added support for custom fields in NaiveThresholdSignatureCircuit;
- Better DLOG keys handling: added possibility to trim keys without regenerating them based on the size of the circuit (https://github.com/HorizenOfficial/zendoo-sc-cryptolib/pull/77);
- Be more verbose when error occurs via eprintln! (will be replaced with Java exceptions and adequate description when addressing #31 

Must change the base branch to development when ready to merge (and previous dependent ones have been merged)


demo-circuit/src/naive_threshold_sig/mod.rs
```diff
@@ -117,19 +119,31 @@ impl<F: PrimeField> NaiveTresholdSignature<F> {
...
-                .update(wcert_sysdata_hash)
-                .finalize()
-                .unwrap()
+            
+            // Compute custom_fields hash taking into account the presence, or not, of custom_fields
+            let to_hash = if let Some(custom_fields) = custom_fields.clone() {
+                let mut h = FieldHash::init_constant_length(custom_fields.len(), None);
+                custom_fields.into_iter().for_each(|custom_field| { h.update(custom_field); });
+                let custom_fields_hash = h.finalize().unwrap();
+                vec![custom_fields_hash, wcert_sysdata_hash]
```
i-Alex, line 139: Why `custom_fields_hash` is not a part of `wcert_sysdata_hash`?
Is it fine and consistent with MC (if any relation)?

jni/src/main/java/com/horizen/sigproofnative/NaiveThresholdSigProof.java
```diff
@@ -229,20 +375,39 @@ public static boolean verifyProof(
...
             long quality,
+            Optional<List<FieldElement>> customFields,
             byte[] proof,
             boolean checkProof,
             String verificationKeyPath,
             boolean checkVerificationKey
     )
     {
+        // Passing an empty array, Rust side, instead of an Option, will slightly simplify the process of unpacking.
+        FieldElement[] customFieldsArray = (customFields.isPresent()) ? customFields.get().toArray(new FieldElement[0]): new FieldElement[] {};
```
i-Alex, line 386: Why not to simplify the interface of `verifyProof` and change `Optional<List<FieldElement>>` to `List<FieldElement>`?
Anyway inside if there are no custom fields we convert Optional it to empty list.

i-Alex (changes requested):

jni/src/main/java/com/horizen/sigproofnative/NaiveThresholdSigProof.java
DanieleDiBenedetto: Conversion to an empty list is performed because otherwise it would be a mess to unpack the Optional Rust side using all the JNI methods and so on. If you feel it's ok to pass an empty list when we don't have any custom field, ok I can modify the interface, no problem

demo-circuit/src/naive_threshold_sig/mod.rs
DanieleDiBenedetto: This reflects exactly the computation performed in cctp-lib (https://github.com/HorizenOfficial/zendoo-cctp-lib/blob/dev/cctp_primitives/src/proving_system/verifier/certificate.rs#L11), so for sure mc-cryptolib and zendoo are aligned in the same way.
We don't have any certificate struct Rust side so it doesn't really matter at the end of the day

jni/src/main/java/com/horizen/sigproofnative/NaiveThresholdSigProof.java
i-Alex: rust side case is clear and fine for me.
The point here was exactly about java method interface and this line.
So, let's change the interface only. Same in similar places.

demo-circuit/src/naive_threshold_sig/mod.rs
i-Alex: great, thanks!

jni/src/main/java/com/horizen/sigproofnative/NaiveThresholdSigProof.java
DanieleDiBenedetto: Will do, thanks !

DanieleDiBenedetto: Done

Yousefatm999 (approved):

jni/src/main/java/com/horizen/cswnative/CswFtProverData.java
```diff
@@ -0,0 +1,99 @@
...
+import com.horizen.librustsidechains.FieldElement;
+import com.horizen.merkletreenative.MerklePath;
+
+public class CswFtProverData implements AutoCloseable {
+    private final FordwardTransferOutput output;
+    private final byte[] ftInputSecretKey;
+    private final FieldElement mcbScTxsComStart;
+    private final MerklePath merklePathToScHash;
+    private final MerklePath ftTreePath;
+    private final FieldElement scCreationTxHash;
```
i-Alex, line 16: let's rename it to `scCreationCommitment`.

jni/src/main/java/com/horizen/cswnative/CswProof.java
```diff
@@ -0,0 +1,342 @@
...
+     * @param compressPk - if the proving key must be saved to provingKeyPath in compressed form
+     * @param compressVk - if the verification key must be saved to verificationKeyPath in compressed form
+     * @return true if (pk, vk) generation and saving to file was successfull, false otherwise
+     */
+    public static boolean setup(
+        ProvingSystemType psType,
+        int rangeSize,
+        int mstTreeHeight,
+        int ftTreeHeight,
+        int scTxCommTreeHeight,
```
i-Alex, line 44: why not to hardcode `scTxCommTreeHeight` and `ftTreeHeight` on rust side getting the values from CCTP commitment tree?

jni/src/main/java/com/horizen/cswnative/CswProof.java
```diff
@@ -0,0 +1,342 @@
...
+     *                 withdraw of a SC utxo instead.
+     * @param provingKeyPath - file path from which reading the proving key
+     * @param checkProvingKey - enable semantic checks on the proving key (WARNING: very expensive)
+     * @param zk - if proof must be created using zk property or not
+     * @param compressed_pk - if the pk read from provingKeyPath is in compressed form or not
+     * @param compress_proof - whether to return the proof bytes in compressed form or not
+     * @return a CreateProofResult instance, i.e. the computed proof bytes and the quality of the certificate (i.e. in this case, number of valid signatures),
+     *         OR null pointer if some errors occured during proof creation.
+     */
+    public static CreateProofResult createProof(
```
i-Alex, line 150: add `throws IllegalArgumentException`. Same in other places.

jni/src/main/java/com/horizen/cswnative/CswFtProverData.java
DanieleDiBenedetto: Done

jni/src/main/java/com/horizen/cswnative/CswProof.java
DanieleDiBenedetto: Done

jni/src/main/java/com/horizen/cswnative/CswProof.java
DanieleDiBenedetto: You're right, it was just to be more generic, but maybe it's not needed. These values are specified in the Constants class anyway.

jni/src/main/java/com/horizen/cswnative/CswProof.java
```diff
@@ -4,13 +4,27 @@
 
 import com.horizen.certnative.WithdrawalCertificate;
 import com.horizen.librustsidechains.FieldElement;
-import com.horizen.provingsystemnative.CreateProofResult;
 import com.horizen.provingsystemnative.ProvingSystemType;
 
 public class CswProof {
+
+    private final int rangeSize;
+    private final int numCustomFields;
```
i-Alex, line 12: honestly, previous structure was better as for me.
I mean, no constructor with `rangeSize` and `numCustomFields`, but mking them an args of `static` methods: setup (...) and proof(...)
PS: we never keep an instance of `CswProof` in SDK, but wrap the static methods in a java friendly way.

DanieleDiBenedetto: I need them for creating the proof that's why I put them as members of the class. Of course, if you are more comfortable, I can just duplicate them in the static prover methods, no problem

demo-circuit/src/pgd_csw/data_structures.rs
```diff
@@ -57,57 +128,74 @@ impl Default for CswUtxoInputData {
...
+    pub wcert_tree_root: FieldElement,   // root hash of the Wcert tree included in ScTxsComm tree
     pub sc_txs_com_hashes: Vec<FieldElement>, // contains all ScTxsComm cumulative hashes on the way from `mcb_sc_txs_com_start` to `mcb_sc_txs_com_end`
                                               // RANGE_SIZE is a number of blocks between `mcb_sc_txs_com_start` and `mcb_sc_txs_com_end`.
                                               // It seems it can be a constant as the number of blocks between the last confirmed block and SC ceasing block should be fixed for a particular sidechain
                                               // witnesses [END]
 }
+
+#[derive(Clone)]
+pub struct CswProverData {
+    // public inputs
```
DanieleDiBenedetto, line 194: Wrong comment, these are used to build the public inputs

demo-circuit/src/pgd_csw/constraints/mod.rs
```diff
@@ -28,15 +29,41 @@ pub mod data_structures;
...
+    ) -> Self {
         CeasedSidechainWithdrawalCircuit {
             sidechain_id,
             csw_data,
+            range_size,
+            num_custom_fields,
         }
     }
+
+    pub fn get_instance_for_setup(_range_size: u32, _num_custom_fields: u32) -> Self {
```
DanieleDiBenedetto, line 64: This function might need to know also whether constant is present or not

demo-circuit/src/blaze_csw/constraints/data_structures.rs
```diff
@@ -0,0 +1,1821 @@
...
+    type FieldGadget = FieldElementGadget;
+
+    fn to_field_gadget_elements<CS: ConstraintSystemAbstract<FieldElement>>(
+        &self,
+        mut cs: CS,
+    ) -> Result<Vec<FieldElementGadget>, SynthesisError> {
+        let mut bits = self.amount_g.to_bits_le();
+        bits.reverse();
+
+        let mut receiver_pub_key_bits = self.receiver_pub_key_g;
```
i-Alex, line 971: There is a misalignment between how mainchain use `receiver_pub_key` for ForwardTransfers and how it is used for proof of public key ownership in `ScPublicKeyGadget::enforce_pk_ownership` inside CSW circuit.

On MC: to send a FT to sidechain we need to provide MC RPC with sc public key hex string generated on SC node.
This public key bytes are transformed into the hex string exactly as it is used by SC noded. Instead, MC reverse that value under the hood, because it is represented as a `uint256`. And this reversed representation is used while computing CommitmentTree.
So, on SDK side pub key is "aabbccdd", on MC is "ddccbbaa".

Then when we want to withdraw FT in a CSW we need to provide FT data exactly how MC/CCTP lib expect it for commitment tree computation. So, to prove that some FT exists in some CommTree `pubKeyBytes` must be passed as "ddccbbaa".

From other side CSW circuit checks the ownership of the public key, but expect it in the original form "aabbccdd".
Because the "receiver_pub_key_g" is a single source of data for both of these checks we will not be able to satisfy both of them inside the circuit.
Note: there is no problem with UTXO withdrawals, since `pubKeyBytes` are used in the expected order.

I would propose to use `pubKeyBytes` in the original order when passing as the witness of CSW circuit and when calculating the nullifier outside the circuit, but reverse the bytes here before calculating the `ft_hash`.

DanieleDiBenedetto: Done

demo-circuit/src/blaze_csw/constraints/data_structures.rs
```diff
@@ -0,0 +1,1833 @@
...
+#[derive(PartialEq, Eq)]
+pub struct CswFtOutputDataGadget {
+    pub amount_g: UInt64,
+    pub receiver_pub_key_g: [UInt8; SC_PUBLIC_KEY_LENGTH],
+    pub payback_addr_data_hash_g: [Boolean; MC_RETURN_ADDRESS_BYTES * 8],
+    pub tx_hash_g: [Boolean; FIELD_SIZE * 8],
+    pub out_idx_g: UInt32,
+}
+
+impl CswFtOutputDataGadget {
```
DanieleDiBenedetto, line 784: We can apply the same optimization as discussed in the previous comment to decide whether it's phantom or not. E.g. let's take payback_addr_data_hash_g as discriminant

demo-circuit/src/blaze_csw/constraints/mod.rs
```diff
@@ -0,0 +1,1652 @@
...
+            range_size,
+            num_custom_fields,
+            sidechain_id: PHANTOM_FIELD_ELEMENT,
+            csw_data: CswProverData::get_phantom(range_size, num_custom_fields),
+            constant: if is_constant_present {
+                Some(PHANTOM_FIELD_ELEMENT)
+            } else {
+                None
+            },
+            csw_sys_data_hash: PHANTOM_FIELD_ELEMENT,
```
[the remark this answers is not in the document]
ptagl: Done with commit [51b8680341bc3a4c82568422d4e6cff927a169d2](https://github.com/HorizenOfficial/zendoo-sc-cryptolib/commit/51b8680341bc3a4c82568422d4e6cff927a169d2).

demo-circuit/src/blaze_csw/data_structures.rs
```diff
@@ -0,0 +1,288 @@
...
+            .unwrap(),
+            quality,
+            mcb_sc_txs_com,
+            ft_min_amount,
+            btr_min_fee,
+            custom_fields,
+        }
+    }
+
+    pub(crate) fn get_phantom(num_custom_fields: u32) -> Self {
```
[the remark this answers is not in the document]
ptagl: Done with commit [51b8680341bc3a4c82568422d4e6cff927a169d2](https://github.com/HorizenOfficial/zendoo-sc-cryptolib/commit/51b8680341bc3a4c82568422d4e6cff927a169d2).

demo-circuit/src/blaze_csw/data_structures.rs
```diff
@@ -0,0 +1,288 @@
...
+        Self {
+            ledger_id: PHANTOM_FIELD_ELEMENT,
+            epoch_id: 0,
+            bt_root: get_bt_merkle_root(None).unwrap(),
+            quality: 0,
+            mcb_sc_txs_com: PHANTOM_FIELD_ELEMENT,
+            ft_min_amount: 0,
+            btr_min_fee: 0,
+            custom_fields: vec![PHANTOM_FIELD_ELEMENT; num_custom_fields as usize],
```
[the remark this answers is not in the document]
ptagl: Since we implemented the `WithdrawalCertificateData::get_default()` function (see commit [51b8680341bc3a4c82568422d4e6cff927a169d2](https://github.com/HorizenOfficial/zendoo-sc-cryptolib/commit/51b8680341bc3a4c82568422d4e6cff927a169d2)), what should be done with `WithdrawalCertificateData::get_phantom()`? Should it just call `get_default()`?

demo-circuit/src/blaze_csw/constraints/mod.rs
```diff
@@ -0,0 +1,1652 @@
...
+    }
+
+    fn compute_mst_tree_data(
+        utxo_output_data: CswUtxoOutputData,
+    ) -> (FieldElement, FieldElement, GingerMHTBinaryPath) {
+        let mut mst = GingerMHT::init(MST_MERKLE_TREE_HEIGHT, 1 << MST_MERKLE_TREE_HEIGHT).unwrap();
+
+        let mst_leaf_hash = utxo_output_data.hash(None).unwrap();
+
+        mst.append(mst_leaf_hash).unwrap();
```
doc78, line 435: I see we are not enforcing in the circuit (and thus not computing here) that the position of the leaf (leaf_index) should be equal to the hash of the output_data mod the total number of leaves of the tree. Daniele told me that the design is a bit different from the original, so I guess it is not needed here, but please double check this.

demo-circuit/src/blaze_csw/data_structures.rs
```diff
@@ -0,0 +1,288 @@
...
+            && self.bt_root == other.bt_root
+            && self.quality == other.quality
+            && self.mcb_sc_txs_com == other.mcb_sc_txs_com
+            && self.ft_min_amount == other.ft_min_amount
+            && self.btr_min_fee == other.btr_min_fee
+            && self.custom_fields == other.custom_fields
+    }
+}
+
+//TODO: Is it ok to define phantom spending_pub_key as an array of all 0s ?
```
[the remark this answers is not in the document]
ptagl: Done with [commit](https://github.com/HorizenOfficial/zendoo-sc-cryptolib/commit/4eee331fd886ced705ed4d127ba5ca09033ccb0f).

@DanieleDiBenedetto should we change also the initialization of the `secret_key` field in `CswUtxoInputData::default()` and `CswUtxoInputData::get_phantom()`?

demo-circuit/src/blaze_csw/data_structures.rs
```diff
@@ -0,0 +1,288 @@
...
+}
+
+#[derive(Clone)]
+pub struct CswUtxoInputData {
+    pub output: CswUtxoOutputData,
+    pub secret_key: [bool; SIMULATED_SCALAR_FIELD_MODULUS_BITS],
+}
+
+impl Default for CswUtxoInputData {
+    fn default() -> Self {
```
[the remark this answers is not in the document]
ptagl: Done with [commit](https://github.com/HorizenOfficial/zendoo-sc-cryptolib/commit/c23a27fe0881b75e5b21d82e6c883f8411ad850c).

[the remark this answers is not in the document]
ptagl: Done with [commit](https://github.com/HorizenOfficial/zendoo-sc-cryptolib/commit/c23a27fe0881b75e5b21d82e6c883f8411ad850c).

demo-circuit/src/blaze_csw/data_structures.rs
```diff
@@ -0,0 +1,288 @@
...
+
+impl PartialEq for CswUtxoInputData {
+    fn eq(&self, other: &Self) -> bool {
+        self.output == other.output && self.secret_key == other.secret_key
+    }
+}
+
+//TODO: Is it ok to define phantom receiver_pub_key as an array of all 0s ?
+#[derive(Clone, Default)]
+pub struct CswFtOutputData {
```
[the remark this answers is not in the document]
ptagl: Done with [commit](https://github.com/HorizenOfficial/zendoo-sc-cryptolib/commit/3926283962323d6abe94e43799e0d98900f5224c).

demo-circuit/src/blaze_csw/data_structures.rs
```diff
@@ -0,0 +1,288 @@
...
+    pub input: CswUtxoInputData, // unspent output we are trying to withdraw
+    pub mst_path_to_output: GingerMHTBinaryPath, // path to output in the MST of the known state
+}
+
+impl Default for CswUtxoProverData {
+    fn default() -> Self {
+        Self {
+            input: CswUtxoInputData::default(),
+            mst_path_to_output: GingerMHTBinaryPath::new(vec![
+                (PHANTOM_FIELD_ELEMENT, false);
```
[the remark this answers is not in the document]
ptagl: Done with [commit](https://github.com/HorizenOfficial/zendoo-sc-cryptolib/commit/34880e40807747b00749f3b3209389b40f12859c).

demo-circuit/src/blaze_csw/data_structures.rs
```diff
@@ -0,0 +1,288 @@
...
+impl PartialEq for CswUtxoProverData {
+    fn eq(&self, other: &Self) -> bool {
+        self.input == other.input && self.mst_path_to_output == other.mst_path_to_output
+    }
+}
+
+#[derive(Clone)]
+pub struct CswFtProverData {
+    pub ft_output: CswFtOutputData, // FT output in the MC block
+    pub ft_input_secret_key: [bool; SIMULATED_SCALAR_FIELD_MODULUS_BITS], // secret key that authorizes ft_input spending.
```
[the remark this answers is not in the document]
ptagl: Done with [commit](https://github.com/HorizenOfficial/zendoo-sc-cryptolib/commit/544ed78505fe85a4e70daf3320344c19eb9bde1f).

demo-circuit/src/blaze_csw/data_structures.rs
```diff
@@ -0,0 +1,288 @@
...
+    pub scb_btr_tree_root: FieldElement, // root hash of the BTR tree included in ScTxsComm tree
+    pub wcert_tree_root: FieldElement,   // root hash of the Wcert tree included in ScTxsComm tree
+    pub sc_txs_com_hashes: Vec<FieldElement>, // contains all ScTxsComm cumulative hashes on the way from `mcb_sc_txs_com_start` to `mcb_sc_txs_com_end`
+                                              // RANGE_SIZE is a number of blocks between `mcb_sc_txs_com_start` and `mcb_sc_txs_com_end`.
+                                              // It seems it can be a constant as the number of blocks between the last confirmed block and SC ceasing block should be fixed for a particular sidechain
+                                              // witnesses [END]
+}
+
+impl CswFtProverData {
+    pub(crate) fn get_phantom(commitment_hashes_number: u32) -> Self {
```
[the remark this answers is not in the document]
ptagl: Done (while working on other comments).

demo-circuit/src/blaze_csw/data_structures.rs
```diff
@@ -0,0 +1,288 @@
...
+#[derive(Clone)]
+pub struct CswProverData {
+    pub sys_data: CswSysData,
+    pub last_wcert: WithdrawalCertificateData, // the last confirmed wcert in the MC
+    pub utxo_data: CswUtxoProverData,
+    pub ft_data: CswFtProverData,
+}
+
+impl CswProverData {
+    pub(crate) fn get_phantom(range_size: u32, num_custom_fields: u32) -> Self {
```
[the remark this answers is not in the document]
ptagl: Done (while working on other comments). The `get_phantom()` function is currently unused.

demo-circuit/src/blaze_csw/constraints/data_structures.rs
```diff
@@ -0,0 +1,1833 @@
...
+    pub custom_fields_g: Vec<FieldElementGadget>,
+}
+
+impl WithdrawalCertificateDataGadget {
+    pub fn is_phantom<CS: ConstraintSystemAbstract<FieldElement>>(
+        &self,
+        mut cs: CS,
+        num_custom_fields: u32,
+    ) -> Result<Boolean, SynthesisError> {
+        let phantom_wcert_g = WithdrawalCertificateDataGadget::from_value(
```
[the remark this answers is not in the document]
ptagl: Done with [commit](https://github.com/HorizenOfficial/zendoo-sc-cryptolib/commit/5360b3dc1703b283c0551e15e84f10c666c9039b).

demo-circuit/src/blaze_csw/constraints/data_structures.rs
```diff
@@ -0,0 +1,1833 @@
...
+#[derive(PartialEq, Eq)]
+pub struct CswUtxoOutputDataGadget {
+    pub spending_pub_key_g: [Boolean; SC_PUBLIC_KEY_LENGTH * 8],
+    pub amount_g: UInt64,
+    pub nonce_g: UInt64,
+    pub custom_hash_g: [Boolean; SC_TX_HASH_LENGTH * 8],
+}
+
+impl CswUtxoOutputDataGadget {
+    pub fn is_phantom<CS: ConstraintSystemAbstract<FieldElement>>(
```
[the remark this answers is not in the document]
ptagl: Done with [commit](https://github.com/HorizenOfficial/zendoo-sc-cryptolib/commit/296d12bb8275a52bb067fe59a0343c1159465320).

demo-circuit/src/blaze_csw/constraints/data_structures.rs
```diff
@@ -0,0 +1,1833 @@
...
+}
+
+#[derive(PartialEq, Eq)]
+pub struct CswUtxoInputDataGadget {
+    pub output_g: CswUtxoOutputDataGadget,
+    pub secret_key_g: [Boolean; SIMULATED_SCALAR_FIELD_MODULUS_BITS],
+}
+
+impl CswUtxoInputDataGadget {
+    pub fn is_phantom<CS: ConstraintSystemAbstract<FieldElement>>(
```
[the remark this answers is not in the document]
ptagl: Done with [commit](https://github.com/HorizenOfficial/zendoo-sc-cryptolib/commit/d969eb75e9dc4e5f94f68a1d1f04289709a8da3c).

demo-circuit/src/blaze_csw/constraints/data_structures.rs
ptagl: Done with [commit](https://github.com/HorizenOfficial/zendoo-sc-cryptolib/commit/5bae454eef88900b13e9ec316cd567ba07d13885).

demo-circuit/src/blaze_csw/data_structures.rs
```diff
@@ -233,23 +288,44 @@ pub struct CswFtProverData {
...
+            scb_btr_tree_root: FieldElement::default(),
+            wcert_tree_root: FieldElement::default(),
+            sc_txs_com_hashes: vec![FieldElement::default(); commitment_hashes_number as usize],
+        }
+    }
+
+    pub(crate) fn get_phantom(commitment_hashes_number: u32) -> Self {
+        Self {
+            ft_output: CswFtOutputData::get_phantom(),
+            ft_input_secret_key: [false; SIMULATED_SCALAR_FIELD_MODULUS_BITS],
```
DanieleDiBenedetto, line 315: Let's set it to an array of all 1s, like you did for the CswUtxoInputData

demo-circuit/src/blaze_csw/data_structures.rs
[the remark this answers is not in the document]
DanieleDiBenedetto: Right, let's remove it

demo-circuit/src/constants/mod.rs
```diff
@@ -174,54 +174,13 @@ pub const PHANTOM_FIELD_ELEMENT: FieldElement = field_new!(
     ])
```
DanieleDiBenedetto, line 174: I don't think we need PHANTOM_FIELD_ELEMENT anymore. Let's replace it with FieldElement::default() in all the places where it's used

demo-circuit/src/blaze_csw/constraints/mod.rs
DanieleDiBenedetto: For now should be fine, as we don't use the bit vector

demo-circuit/src/blaze_csw/constraints/mod.rs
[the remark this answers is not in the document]
DanieleDiBenedetto: Right, let's get rid of it

demo-circuit/src/blaze_csw/data_structures.rs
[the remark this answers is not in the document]
DanieleDiBenedetto: Yes, it should be fine

demo-circuit/src/blaze_csw/constraints/mod.rs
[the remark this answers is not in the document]
ptagl: Done with commit [cd72c64b705c3fc92f5fe40a8e540bbdeb9d523f](https://github.com/HorizenOfficial/zendoo-sc-cryptolib/commit/cd72c64b705c3fc92f5fe40a8e540bbdeb9d523f).

demo-circuit/src/blaze_csw/data_structures.rs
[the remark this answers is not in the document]
ptagl: Done with commit [cd72c64b705c3fc92f5fe40a8e540bbdeb9d523f](https://github.com/HorizenOfficial/zendoo-sc-cryptolib/commit/cd72c64b705c3fc92f5fe40a8e540bbdeb9d523f).

demo-circuit/src/blaze_csw/data_structures.rs
ptagl: Done with commit [cd72c64b705c3fc92f5fe40a8e540bbdeb9d523f](https://github.com/HorizenOfficial/zendoo-sc-cryptolib/commit/cd72c64b705c3fc92f5fe40a8e540bbdeb9d523f).

demo-circuit/src/constants/mod.rs
ptagl: Done with commit [cd72c64b705c3fc92f5fe40a8e540bbdeb9d523f](https://github.com/HorizenOfficial/zendoo-sc-cryptolib/commit/cd72c64b705c3fc92f5fe40a8e540bbdeb9d523f).

demo-circuit/src/blaze_csw/constraints/data_structures.rs
```diff
@@ -1070,7 +1069,7 @@ impl CswFtProverDataGadget {
         );
 
         // Alloc phantom field element
-        let phantom_g = FieldElementGadget::from_value(cs.ns(|| "Break"), &PHANTOM_FIELD_ELEMENT);
+        let phantom_g = FieldElementGadget::from_value(cs.ns(|| "Break"), &FieldElement::default());
```
DanieleDiBenedetto, line 1072: Where is this used ?

demo-circuit/src/blaze_csw/constraints/mod.rs
```diff
@@ -0,0 +1,1652 @@
...
+                ck_g1.as_ref().unwrap(),
+                current_public_inputs.as_slice(),
+                &proof
+            )
+            .unwrap());
+
+            // Change one public input each time and check that the proof fails
+            for i in 0..current_public_inputs.len() {
+                let mut wrong_public_inputs = current_public_inputs.clone();
+                wrong_public_inputs[i].double_in_place();
```
doc78, line 801: Actually, double_in_place will not generate a wrong public input if the value of the original one is zero, like a sidechain_id

demo-circuit/src/blaze_csw/constraints/data_structures.rs
ptagl: @DanieleDiBenedetto if you mean the phantom gadget, it is used to handle lists of commitment hashes whose size is less then the maximum size:

```
// Ignore NULL hashes
let should_ignore_hash = self.sc_txs_com_hashes_g[i].is_eq(
    cs.ns(|| format!("sc_txs_com_hashes[{}] == PHANTOM", i)),
    &phantom_g,
)?;
```

demo-circuit/src/blaze_csw/constraints/mod.rs
ptagl: Switched to `add_assign()` with commit [00d8a249d159deeace576b91b1b1c1f895428f3f](https://github.com/HorizenOfficial/zendoo-sc-cryptolib/commit/00d8a249d159deeace576b91b1b1c1f895428f3f).

demo-circuit/src/naive_threshold_sig/mod.rs
```diff
@@ -230,11 +276,24 @@ impl<F: PrimeField> ConstraintSynthesizer<FieldElement> for NaiveTresholdSignatu
             },
         )?;
 
+        // Alloc custom_fields and enforce their hash, if they are present
+        let custom_fields_hash_g = if let Some(custom_fields) = self.custom_fields.clone() {
```
dkaidalov, line 280: If I understand correctly this is a circuit for the BFT Withdrawal Certificate proof. Note that according to specification BFT Wcert contains 2 custom fields: `scb_new_mst_root` and `submitter_key`. And `submitter_key` should not be included in the message signed by the servers, because at the moment of signing the submitter isn't known yet. Here it seems all custom fields constitutes message to sign.

demo-circuit/src/blaze_csw/constraints/mod.rs
```diff
@@ -0,0 +1,1680 @@
...
+            &csw_data_g.sys_data_g.amount_g,
+            &should_enforce_ft_withdrawal_g,
+        )?;
+
+        // We check the public key ownership just once for both, choosing the appropriate public key
+        // and secret key, as it is an expensive check, we want to do it just once.
+        // NOTE: We could've done the same for nullifier and amount checks, but we didn't in order
+        //       to have cleaner code (we lose only 2 constraints anyway)
+
+        ScPublicKeyGadget::enforce_pk_ownership(
```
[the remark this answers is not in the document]
DanieleDiBenedetto: In reality they can be both phantom or even not phantom, previously there was a check enforcing EXACTLY one of the two not being phantom (enforce_xor(should_enforce_utxo_withdrawal, should_enforce_ft_withdrawal). But after a discussion with Oleksandr we agreed that this constraint is useless as the prover cannot supply valid public inputs in either of the cases. If you have doubts about it we can reintroduce that constraint

demo-circuit/src/naive_threshold_sig/mod.rs
DanieleDiBenedetto: Yes, due to an issue in MC we were forced to split a single field element, the scb_new_mst_root, into 2.
submitter_key is not part of the custom fields as I understood.

demo-circuit/src/blaze_csw/constraints/mod.rs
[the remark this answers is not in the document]
dkaidalov: What do you mean they both (utxo and FT) can be phantom? Then how private key verification would pass?
My understanding is that private key verification ensures that one of them is not phantom. Isn't that correct?
Moreover, they can not both be not phantom and pass verification because nullifier cannot be the same for both UTXO and FT. So exactly one of them must be phantom to avoid conditional check. Where am I wrong?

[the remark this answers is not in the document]
DanieleDiBenedetto: Yes you got it perfectly, that's why the XOR check would've been redundant

demo-circuit/src/naive_threshold_sig/mod.rs
dkaidalov: Then, where is the submitter key? Or this circuit is not fully implemented yet?

DanieleDiBenedetto: submitter_pubkey is not used in the circuit, but all custom fields of the certificate are allocated and hashed inside the circuit. I guess it is correctly passed in the SDK, right @i-Alex ?

demo-circuit/src/blaze_csw/constraints/mod.rs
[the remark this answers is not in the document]
DanieleDiBenedetto: Anyway we can add it, just to be on the safe side, it's just 1 constraint anyway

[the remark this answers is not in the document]
dkaidalov: I also do not see a strong reason to have this check, just wanted to confirm that I understood everything correctly.

demo-circuit/src/blaze_csw/mod.rs
```diff
@@ -0,0 +1,326 @@
...
+    let x_sign = if sw_pk.x.is_odd() { 1 << 7 } else { 0u8 };
+
+    // Extract the public key bytes as Y coordinate
+    let y_coordinate = sw_pk.y;
+    let mut pk_bytes = serialize_to_buffer(&y_coordinate, None).unwrap();
+
+    // Use the last (null) bit of the public key to store the sign of the X coordinate
+    // Before this operation, the last bit of the public key (Y coordinate) is always 0 due to the field modulus
+    let len = pk_bytes.len();
+    pk_bytes[len - 1] |= x_sign;
```
UlrichHaboeck75, line 190: This is non-injective serialization: Weierstrass curves can have up to three points  for a given y-value.
I propose to use serialize via x-coordinate, plus the parity bit for y.

demo-circuit/src/blaze_csw/mod.rs
```diff
@@ -0,0 +1,326 @@
...
+
+    Ok(pk_bytes.try_into().unwrap())
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use serial_test::*;
+
+    #[serial]
```
UlrichHaboeck75, line 200: Why is such an elementary test enforced to run in serial?

demo-circuit/src/blaze_csw/constraints/mod.rs
```diff
@@ -0,0 +1,1677 @@
...
+
+            let mut receiver_g_bits = csw_data_g.sys_data_g.receiver_g;
+            receiver_g_bits.reverse();
+
+            amount_and_receiver_bits_g.extend_from_slice(&receiver_g_bits[..]);
+
+            FieldElementGadget::from_bits(
+                cs.ns(|| "read field element out of amount and bits"),
+                amount_and_receiver_bits_g.as_slice(),
+            )
```
UlrichHaboeck75, line 312: It seems to me that we implicitly assume that the length of `amount_and_receiver_bits_g` does not exceed the capacity of the constraint field. For sure this is the case in our bn_382 instantiation, but it might be worth to put an `assert!` here.

demo-circuit/src/blaze_csw/constraints/data_structures.rs
```diff
@@ -0,0 +1,1838 @@
...
+                &(<SimulatedCurveParameters as MontgomeryModelParameters>::COEFF_B
+                    * SimulatedFieldElement::from(3u8)),
+            )?;
+
+            sw_pk_x_coordinate_g.mul_equals(
+                cs.ns(|| "check x_sw"),
+                &three_b_times_one_minus_y_ed,
+                &rhs,
+            )?;
+        }
```
UlrichHaboeck75, line 1552: Although conversion does not take too many constraints in comparison to the scalar multiplication circuit, we can improve here using a single non-native product: Multiplying the transformation equality
```
	x_sw * (1 - y_ed) = 1/B (1 + y_ed) + A/3B (1 - y_ed)
```
by minus one yields
```
	x_sw * (y_ed - 1) = -1/B (y_ed + 1) + A/3B (y_ed - 1).
```
Replacing `y_ed + 1` by `y_ed - 1  + 2`, and gathering the `y_ed - 1` terms results in
the single product equality 
```
	(x_sw + 1/B - A/3B) * (y_ed - 1) = -2/B.
```

demo-circuit/src/blaze_csw/mod.rs
DanieleDiBenedetto: Yes you're right, it's not used but I changed it

demo-circuit/src/blaze_csw/mod.rs
DanieleDiBenedetto: No specific reason. The greatest part of the tests are memory intensive so I run all of them serially anyway

demo-circuit/src/blaze_csw/constraints/data_structures.rs
DanieleDiBenedetto: Done

demo-circuit/src/blaze_csw/constraints/mod.rs
```diff
@@ -0,0 +1,1682 @@
...
+    ) -> Result<(), SynthesisError> {
+        // Alloc sidechain id
+        let sidechain_id_g =
+            FieldElementGadget::alloc(cs.ns(|| "alloc sidechain_id_g"), || Ok(&self.sidechain_id))?;
+
+        // Alloc all witness data
+        let csw_data_g =
+            CswProverDataGadget::alloc(cs.ns(|| "alloc csw data"), || Ok(&self.csw_data))?;
+
+        // Decide whether to enforce utxo or ft withdrawal
```
UlrichHaboeck75, line 187: When reading it the first time, the independent choice of the the `should_enforc_utxo/ft_withdrawal_g` made me wonder. 
Maybe it is worth to mention in a comment that it is as good as impossible to satisfy both a utxo and a ft withdrawal within this circuit.

demo-circuit/src/blaze_csw/constraints/mod.rs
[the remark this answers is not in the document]
UlrichHaboeck75: My five cents on this: Oleksandr's argument is perfeclty fine. However, a clarification of this in a comment would be helpful (it was giving me a hard time when reading the code).

demo-circuit/src/constants/mod.rs
```diff
@@ -154,13 +154,32 @@ impl VRFParams {
...
-    use super::*;
     use algebra::{AffineCurve, FpParameters, FromCompressedBits, PrimeField};
+    use cctp_primitives::utils::serialization::serialize_to_buffer;
+
+    use super::*;
     use bit_vec::BitVec;
     use blake2s_simd::{Hash, Params};
 
+    use serial_test::*;
+
```
UlrichHaboeck75, line 182: Not related to Line 182, but the `fn hash_to_curve()`: 
Why is not the more performant "hunting and pecking" strategy used? That one simple computes a starting x-coordinate via hash, and then increment it until we find a point for it.

demo-circuit/src/constants/mod.rs
```diff
@@ -266,4 +287,42 @@ mod test {
...
+            5404726604943382876,
+        ]));
+        let y = SimulatedFieldElement::from(BigInteger256([
+            3725370832501899225,
+            12695286482625208691,
+            7386704395789842856,
+            3479569230309726808,
+        ]));
+        let simulated_te_point = SimulatedTEGroup::new(x, y);
+        assert_eq!(simulated_te_point.group_membership_test(), false);
```
UlrichHaboeck75, line 307: Actually, it is not obvious that our `hash_to_curve` outputs a point outside the prime order subgroup, given the specific `Strontium` string. I would remove this assertion, as it does not play any role for security.

demo-circuit/src/blaze_csw/constraints/mod.rs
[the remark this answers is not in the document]
DanieleDiBenedetto: Added a comment about this

demo-circuit/src/blaze_csw/constraints/mod.rs
DanieleDiBenedetto: Added a comment about this

demo-circuit/src/blaze_csw/constraints/data_structures.rs
```diff
@@ -0,0 +1,1833 @@
...
+    }
+
+
+    fn convert_te_to_sw_point<CS: ConstraintSystemAbstract<FieldElement>>(
+        mut cs: CS,
+        te_pk_x_coordinate_g: NonNativeFieldGadget<SimulatedFieldElement, FieldElement>,
+        te_pk_y_coordinate_g: NonNativeFieldGadget<SimulatedFieldElement, FieldElement>,
+    ) -> Result<ECPointSimulationGadget, SynthesisError> {
+        let y_ed_minus_one = te_pk_y_coordinate_g
+            .add_constant(cs.ns(|| "y_ed - 1"), &-SimulatedFieldElement::one())?;
```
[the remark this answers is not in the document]
DanieleDiBenedetto: Added a comment abouit this

demo-circuit/src/constants/mod.rs
DanieleDiBenedetto: Well that's what we do it basically but instead of incrementing the x_coordinate, we increment an internal counter that produces a new hash

demo-circuit/src/blaze_csw/constraints/mod.rs
```diff
@@ -184,7 +198,11 @@ impl ConstraintSynthesizer<FieldElement> for CeasedSidechainWithdrawalCircuit {
         let csw_data_g =
             CswProverDataGadget::alloc(cs.ns(|| "alloc csw data"), || Ok(&self.csw_data))?;
 
-        // Decide whether to enforce utxo or ft withdrawal
+        // Decide whether to enforce utxo or ft withdrawal.
+        // NOTE: We don't need to explicitly enforce that exactly one among 'should_enforce_utxo_withdrawal_g'
+        //       and 'should_enforce_ft_withdrawal_g' must be True (doable with a XOR constraint) as:
+        //       - They cannot be both phantom as pk ownership check won't pass;
```
UlrichHaboeck75, line 204: It might happen that our choice of the null key even allows an ownership proof. It is the collision resistance of the nullifier  that at most one of the cases (utxo or ft) can be satisified.

UlrichHaboeck75 (approved):
Approve the code in `blaze_csw/`.

demo-circuit/src/blaze_csw/constraints/mod.rs
```diff
@@ -190,49 +143,73 @@ impl ConstraintSynthesizer<FieldElement> for CeasedSidechainWithdrawalCircuit {
...
+
+            (
+                CswProverDataGadget::alloc(cs.ns(|| "alloc csw data"), || Ok(csw_data))?,
+                actual_range_size,
+            )
+        };
 
         // if last_wcert != NULL
         // enforce(sys_data.sc_last_wcert_hash == H(last_wcert))
+        // TODO: Do we need to enforce this also in case of FT withdrawal ?
```
DanieleDiBenedetto, line 198: @UlrichHaboeck75 @dkaidalov @albertog78 we should understand this

demo-circuit/src/blaze_csw/constraints/data_structures.rs
```diff
@@ -0,0 +1,1496 @@
...
+            .output_g
+            .to_field_gadget_elements(cs.ns(|| "alloc output hash elements"))?;
+
+        debug_assert_eq!(output_hash_elements_g.len(), 3);
+        output_hash_elements_g.push(box_type_coin_g);
+
+        let output_hash_g = FieldHashGadget::enforce_hash_constant_length(
+            cs.ns(|| "H(input.output)"),
+            &output_hash_elements_g,
+        )?;
```
UlrichHaboeck75, line 422: As described in [snarks-architecture/issues/3](https://github.com/HorizenLabs/snarks-architecture/issues/3), the current approach to separate utxo/ft nullifiers is a little bit "dangerous", as its security depends on the type of data to be hashed. Let us use Poseidon's personalization feature to circumvent this. A choice of different personalization string/seeds guarantees that the resulting hash functions used for computing nullifiers are ensured to be uncorrelated. (The same is true if personalization is used for only one of the hashes). As domain separation is a useful feature for a transparent design of cryptographic protocols in general, I propose to

- implement a personalization gadget for Poseidon, which directly sets the internal state to the result of the personalization function from the primitive. This gadget does not cost any constraints.

Then, apply 
-  Poseidon with personalization string such as `seed = "csw_utxo"` for computing the nullifier of utxos
`nullifier = PH_seed(input.output)`, and 
- Poseidon without personalization for the nullifier of forward transactions, `nullifier = PH(ft_input)`.

UlrichHaboeck75 (commented):
I have added some inline docu for blaze_csw. 
Once we clarify the code [optimization issue](https://github.com/HorizenLabs/snarks-architecture/issues/4) approvable.

demo-circuit/src/blaze_csw/constraints/mod.rs
UlrichHaboeck75: @dkaidalov I have posted also a [question/issue](https://github.com/HorizenLabs/snarks-architecture/issues/4) about that in the snark-architecture repo.

demo-circuit/src/naive_threshold_sig/mod.rs
dkaidalov: Yes, everything you described is correct. Some custom fields should be signed (scb_new_mst_root) and some other should not (submitter_key). At some point this should be implemented (at least for full Latus sidechain). But for the BFT we will not have submitter_key, did I understand correctly?

demo-circuit/src/blaze_csw/constraints/mod.rs
dkaidalov: In general I agree with the raised issue in snark-architecture repo: we don't need to enforce last_wcert hash for FT, thus it is better to modify the specification as suggested for better readability. 
But note that in this case `last_wcert` witness is not checked at all in case of FT withdrawal. is it ok? Isn't there any way to abuse proving system in this case? Wasn't that the initial reason why `last_wcert` hash check was moved outside of FT verification branch?

DanieleDiBenedetto: Moving outside it doesn't really solve the problem, as the prover has still freedom to choose whether to enforce or not the cert hash verification (and we are obliged to give him this possibility as a last wcert might not be present when withdrawing a FT). We discussed this with @albertog78 too and our proposal would be to:

1) Like @UlrichHaboeck75 suggested in the issue, enforce this check being performed only when a utxo is withdrawn;
2) In all the other cases, we adopt the same strategy as for `constant` and `receiver ` : the prover is obliged to supply as public input the correct `sys_data.sc_last_wcert_hash` as the verifier would do, otherwise the proof verification will fail.

Other solutions:

- Decide whether to enforce certificate verification or not based on the value of the public input itself being phantom or not (note that in case of FT, this would be exactly equal to the previous solution);
- If we are withdrawing a FT and we don't care about the certificate, then it's ok to give freedom to the prover to decide whether to enforce cert hash or not.

dkaidalov: Moving this check outside reduces options of the prover to: either pass NULL (phantom) or valid last_wcert. 
If we check only inside FT branch then the prover has freedom to provide any random data as last_wcert - this is my concert, can it be exploited somehow?
If not, then I'm fine with Ulrich suggestion. The 2 option is also good.

DanieleDiBenedetto: If certificate being present or not doesn't matter when withdrawing a FT (and it seems so as it's not used) then I don't think it can be exploited in a malicious way anyway

DanieleDiBenedetto: Actually, currently, solutions 1 and 2 are already implemented

DanieleDiBenedetto: Actually, it's already implemented in the way we proposed

dkaidalov: So, as I see the first approach is implemented in the code (check last_wcert only in case UTXO withdrawal), right? Then, I will update the spec accordingly.

DanieleDiBenedetto: Sorry, maybe there is a misunderstanding: 1) and 2) are not 2 distinct approaches but 2 steps of the same one ! So we enforce the cert_hash only in UTXO case and we use the non-malleability property to ensure the prover passes nevertheless the same correct certificate hash as the verifier when creating the proof. Currently everything is already implemented in the code

dkaidalov: ok, got it. Will update the spec accordingly

DanieleDiBenedetto: I added a couple of UTs stressing this case in https://github.com/HorizenOfficial/zendoo-sc-cryptolib/pull/79/commits/54068c4445698f27777d8875880331726dfd20bf if you want to take a look at it

PR #74, opened 2021-11-25
CommitmentTree interface updated.

i-Alex:


api/src/lib.rs
```diff
@@ -2600,6 +2600,126 @@ ffi_export!(
...
 });
 
+ffi_export!(
+    fn Java_com_horizen_commitmenttree_CommitmentTree_nativeGetFwtLeaves(
+    _env: JNIEnv,
+    _commitment_tree: JObject,
+    _sc_id: jbyteArray
+) -> jobject
+{
+    let sc_id = {
```
DanieleDiBenedetto, line 2610: You can use the "deserialize_to_jobject" function to save a little bit of code lines and duplication

api/src/lib.rs
```diff
@@ -2600,6 +2600,126 @@ ffi_export!(
...
+        read_mut_raw_pointer(&_env, t.j().unwrap() as *mut CommitmentTree)
+    };
+
+    match commitment_tree.get_fwt_leaves(&sc_id) {
+        Some(leaves) => {
+            let field_class =  _env.find_class("com/horizen/librustsidechains/FieldElement")
+                .expect("Should be able to find FieldElement class");
+
+            let initial_element = _env.new_object(field_class, "(J)V", &[
+                JValue::Long(0)]).expect("Should be able to create new long for FieldElement");
```
DanieleDiBenedetto, line 2629: In these cases, you can use the "return_jobject" function to save a little bit of code lines and duplication. But don't bother, we will highly refactor the library anyway

api/src/lib.rs
```diff
@@ -2600,6 +2600,126 @@ ffi_export!(
...
+                    .expect("Should be able to add FieldElement leaf to an array");
+            }
+
+
+            let cls_optional = _env.find_class("java/util/Optional").unwrap();
+
+            let empty_res = _env.call_static_method(cls_optional, "of", "(Ljava/lang/Object;)Ljava/util/Optional;", &[JValue::from(JObject::from(leaf_fe_array))])
+                .expect("Should be able to create new value for Optional");
+
+            *empty_res.l().unwrap()
```
DanieleDiBenedetto, line 2650: Again, not important, to save a little bit of code we might consider writing and using a generic function that wraps a JObject into another Optional JObject

DanieleDiBenedetto (approved):

api/src/lib.rs
i-Alex: Thanks!
I completely agree with you. In these methods there is a lot of boilerplate that should be moved to common utility methods.
Currently, just reused the same approach as we have in other places.

albertog78 (approved):

PR #76, opened 2021-12-28
CSW circuit debug feature

i-Alex:
Just a simple update to the interface to be able to detect failing constraint during CSW proof creation. Was very helpful on SDK side during integration.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -44,6 +44,7 @@ dependencies = [
  "jni",
  "lazy_static",
  "primitives",
+ "r1cs-core",
  "radix_trie",
  "rand",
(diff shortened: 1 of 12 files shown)


ptagl (approved):

DanieleDiBenedetto (approved):
@i-Alex isn't it a better to add a separate method in Java and a separate function in Rust returning a String instead of doing everything inside the create proof and simply printing if there is any failing constraint ?

i-Alex:
I agree. Would be nice to have a separate debug circuit method even for threshold signature circuit

i-Alex (commented):
Looks good!

Issue #89, opened 2022-07-22
Add possibility to derive VRF and Schnorr keys from seed


DanieleDiBenedetto:

i-Alex:
The same for Schnorr keys

DanieleDiBenedetto:
Thanks, title fixed

PR #91, opened 2022-08-31
fixes issue 89

mikelodder7:
Address issue #89

### DIFF
--- a/api/src/cctp_calls.rs
+++ b/api/src/cctp_calls.rs
@@ -1,4 +1,5 @@
 use algebra::{AffineCurve, ProjectiveCurve, ToConstraintField, UniformRand};
+use blake2::digest::{FixedOutput, Input};
 use demo_circuit::{
     constants::VRFParams, constraints::CeasedSidechainWithdrawalCircuit, naive_threshold_sig::*,
     type_mapping::*, CswFtProverData, CswSysData, CswUtxoProverData, WithdrawalCertificateData,
@@ -10,7 +11,8 @@ use primitives::{
     vrf::{ecvrf::FieldBasedEcVrfPk, FieldBasedVrf},
 };
 use r1cs_core::debug_circuit;
-use rand::{rngs::OsRng, SeedableRng};
+use rand::{rngs::OsRng, CryptoRng, RngCore, SeedableRng};
+use rand_chacha::ChaChaRng;
 use rand_xorshift::XorShiftRng;
 
 use cctp_primitives::{
@@ -29,6 +31,7 @@ use cctp_primitives::{
     },
 };
 
+use std::convert::TryFrom;
 use std::path::Path;
 
 //*******************************Generic functions**********************************************
@@ -44,8 +47,7 @@ pub fn get_random_field_element(seed: u64) -> FieldElement {
 
 pub fn schnorr_generate_key() -> (SchnorrPk, SchnorrSk) {
     let mut rng = OsRng;
-    let (pk, sk) = SchnorrSigScheme::keygen(&mut rng);
-    (pk.0.into_affine(), sk)
+    schnorr_generate_key_from_rng(&mut rng)
 }
 
 pub fn schnorr_get_public_key(sk: &SchnorrSk) -> SchnorrPk {
@@ -76,6 +78,32 @@ pub fn schnorr_verify_signature(
     signature: &SchnorrSig,
 ) -> Result<bool, Error> {
     SchnorrSigScheme::verify(&FieldBasedSchnorrPk(pk.into_projective()), *msg, signature)
+}
+
+pub fn schnorr_derive_key_from_seed(seed: &[u8]) -> (SchnorrPk, SchnorrSk) {
+    // zero just default to random,
+    // however, is there a minimum length that should be required?
+    if seed.len() == 0 {
+        return schnorr_generate_key();
+    }
+
+    // Domain separation tag
+    const DST: &[u8] = &[0xFFu8; 32];
+
+    // Hash first to ensure size an eliminate any bias
+    // that may exist in `seed`
+    let mut hasher = blake2::Blake2b::default();
+    hasher.input(DST);
+    hasher.input(seed);
+    let digest = hasher.fixed_result();
+    let rng_seed = <[u8; 32]>::try_from(&digest[..32]).unwrap();
+    let mut rng = ChaChaRng::from_seed(rng_seed);
+    schnorr_generate_key_from_rng(&mut rng)
+}
+
+fn schnorr_generate_key_from_rng<R: RngCore + CryptoRng>(rng: &mut R) -> (SchnorrPk, SchnorrSk) {
+    let (pk, sk) = SchnorrSigScheme::keygen(rng);
+    (pk.0.into_affine(), sk)
 }
 
 //*****************************Naive threshold sig circuit related functions************************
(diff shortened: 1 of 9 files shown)


api/src/cctp_calls.rs
```diff
@@ -78,6 +80,32 @@ pub fn schnorr_verify_signature(
...
+    const DST: &[u8] = &[0xFFu8; 32];
+
+    // Hash first to ensure size an eliminate any bias
+    // that may exist in `seed`
+    let mut hasher = blake2::Blake2b::default();
+    hasher.input(DST);
+    hasher.input(seed);
+    let digest = hasher.fixed_result();
+    let rng_seed = <[u8; 32]>::try_from(&digest[..32]).unwrap();
+    let mut rng = ChaChaRng::from_seed(rng_seed);
```
DanieleDiBenedetto, line 100: Is ChaChaRng secure enough ? Quoting from crates.io:

> The ChaCha algorithm is widely accepted as suitable for cryptographic purposes, **but this implementation has not been verified as such**. Prefer a generator like OsRng that defers to the operating system for cases that need high security.

jni/src/main/java/com/horizen/schnorrnative/SchnorrKeyPair.java
```diff
@@ -21,6 +21,12 @@ public SchnorrKeyPair(SchnorrSecretKey secretKey) {
         this.publicKey = secretKey.getPublicKey();
     }
 
+    private static native SchnorrKeyPair nativeDeriveFromSeed(byte[] seed);
+
+    public static SchnorrKeyPair deriveFromSeed(byte[] seed) {
```
DanieleDiBenedetto, line 26: Minor thing: We can just overload the generate() function by passing the additional seed parameter instead of using a different name

jni/src/main/java/com/horizen/vrfnative/VRFKeyPair.java
```diff
@@ -28,6 +28,12 @@ public static VRFKeyPair generate() {
         return nativeGenerate();
     }
 
+    private static native VRFKeyPair nativeDeriveFromSeed(byte[] seed);
```
DanieleDiBenedetto, line 31: Same comment as in the SchnorrKeyPair class

jni/src/test/java/com/horizen/schnorrnative/SchnorrKeyPairTest.java
```diff
@@ -17,6 +17,22 @@ public void testGenerate() throws Exception {
         }
     }
 
+    @Test
+    public void testDeriveFromSeed() throws Exception {
+        byte[] seed = { 1, 2, 3, 4, 5, 6, 7, 8 };
```
DanieleDiBenedetto, line 22: Can we hardcode also (pk, sk) and check the code generates the same ?

jni/src/test/java/com/horizen/vrfnative/VRFKeyPairTest.java
```diff
@@ -21,6 +21,22 @@ public void testGenerate() throws Exception {
         }
     }
 
+    @Test
+    public void testDeriveFromSeed() throws Exception {
+        byte[] seed = { 1, 2, 3, 4, 5, 6, 7, 8 };
```
DanieleDiBenedetto, line 26: Same comment as in the Schnorr test

DanieleDiBenedetto (changes requested):

DanieleDiBenedetto:
Must also fix the CI, it seems that some dependency is not compatible with the hardcoded Rust version

api/src/cctp_calls.rs
mikelodder7: Its used by the Linux Kernel which is periodically seeded from the kernel entropy pool. The only difference here is the seed vs using OsRng which uses ChaCha from /dev/random. We can yell at the caller with comments about how the seed should be long enough like 32 bytes.

jni/src/main/java/com/horizen/schnorrnative/SchnorrKeyPair.java
mikelodder7: That would require an API breaking change. If we are okay with this I can make that change.

jni/src/test/java/com/horizen/schnorrnative/SchnorrKeyPairTest.java
mikelodder7: Sure will add.

mikelodder7:
> Must also fix the CI, it seems that some dependency is not compatible with the hardcoded Rust version

Which is odd because the only thing I added was rand_chacha. Its complaining about the hashbrown crate.

jni/src/main/java/com/horizen/schnorrnative/SchnorrKeyPair.java
DanieleDiBenedetto: Why ? Just add a separate generate(byte[] seed) function

DanieleDiBenedetto: Ah right because Rust side we don't have the overload. You could manage it by adding a single native function that always take a seed, and let the `generate(void)` call that one passing an empty seed (that you already handled Rust side like a normal generate without seed). Btw is not so important

DanieleDiBenedetto (approved):

api/src/cctp_calls.rs
```diff
@@ -805,6 +870,29 @@ mod test {
...
+    #[serial]
+    #[test]
+    fn sample_calls_schnorr_vrf_derive_different_keys_from_same_seed() {
+        const SEED: [u8; 32] = [7u8; 32];
+        let (spk, ssk) = schnorr_derive_key_from_seed(&SEED); //Keygen
+        let (vpk, vsk) = vrf_derive_key_from_seed(&SEED); //Keygen
+
+        assert_ne!(vpk, spk);
+        assert_ne!(vsk, ssk);
+    }
```
i-Alex, line 894: Let's add some regression tests for both schnorr and vrf keys as it was done in the Java part.
It seems more safe and useful, considering that only rust part `sc-sctyprolib` can be used by third parties. So they will be sure that the algorithm remains the same.

jni/src/test/java/com/horizen/schnorrnative/SchnorrKeyPairTest.java
```diff
@@ -17,6 +17,36 @@ public void testGenerate() throws Exception {
...
+        byte[] emptySeed = {};
+        try(SchnorrKeyPair keyPair = SchnorrKeyPair.generate(seed))
+        {
+            assertNotNull("Key pair derive from empty seed was unsuccessful", keyPair);
+            assertTrue("Public key verification failed.", keyPair.getPublicKey().verifyKey());
+        }
+
+    }
+
+    private void assertArrayEquals(String message, byte[] expected, byte[] actual) {
```
i-Alex, line 41: `assertArrayEquals` is a part of JUnit suite. 
Can be reused.
same in other test

mikelodder7: Except that it isn't. I thought so as well but then travis-CI couldn't compile it so I wrote it myself. See [Job 581609520](https://app.travis-ci.com/github/HorizenOfficial/zendoo-sc-cryptolib/jobs/581609520)

api/src/cctp_calls.rs
mikelodder7: Added.

jni/src/test/java/com/horizen/schnorrnative/SchnorrKeyPairTest.java
DanieleDiBenedetto: Perhaps we are using an old version of JUnit ?
If not let's keep it, but can you move it please to the TestUtils class ?

i-Alex (approved):

PR #96, opened 2022-11-29
Check signed tag function fix

ysibirski:
The current bug is not affecting this CI/CD but it needs to be fixed anyway.


otoumas (approved):
LGTM

Issue #97, opened 2022-11-30
Circuits' regression tests for backward compatibility


DanieleDiBenedetto:
When we add new circuits, and/or modifying existing ones, we need to be sure to not break backward compatibility.
It might be helpful to add regression tests, once fixed all external and circuits' parameters, asserting that new circuits' keys are the same as the old ones.

cc: @i-Alex

PR #98, opened 2022-12-02
Restored backward compatibility on ThresholdV1 + added related regression tests

DanieleDiBenedetto:
We noticed a difference in the (pk, vk) pair for ThresholdV1 after the 0.6.0-SNAPSHOT.
This is due to the fact that, accidentally, Poseidon Hash gadget was replaced with an optimized version.
This PR restores backward compatibility, also introducing regression tests to check that such accidental changes could be identified in the future (see issue #97 )

Also an outdated test file was removed (accidentaly present due to merging/rebasing issues I guess)


demo-circuit/src/type_mapping.rs
```diff
@@ -45,12 +45,17 @@ pub type VRFScheme = FieldBasedEcVrf<FieldElement, G2Projective, FieldHash, Grou
 pub type VRFProof = FieldBasedEcVrfProof<FieldElement, G2Projective>;
 pub type VRFPk = G2;
 pub type VRFSk = ScalarFieldElement;
-pub type FieldElementGadget = FpGadget<FieldElement>;
-pub type FieldHashGadget = TweedleFrDensityOptimizedPoseidonHashGadget;
+pub(crate) type FieldElementGadget = FpGadget<FieldElement>;
+pub(crate) type FieldHashGadget = TweedleFrDensityOptimizedPoseidonHashGadget;
+
+// Kept for backward compatibility of Threshold Circuit V1
+pub(crate) type UnoptimizedFieldHashGadget = TweedleFrPoseidonHashGadget;
```
nicholas-mainardi, line 52: Does it make sense to introduce this type? Can't Threshold Circuit V1 directly employ the `TweedleFrPoseidonHashGadget` as in the previous version?

nicholas-mainardi (approved):

demo-circuit/src/type_mapping.rs
DanieleDiBenedetto: I did it in order to make it explicit, so we have less chance to be confused next time D:

PR #100, opened 2023-01-05
Fix Potential Underflows When Converting 64-bit Integers to Field Elements

nicholas-mainardi:
Fix for the issue described in [PS 205](https://horizenlabs.atlassian.net/browse/PS-205):

When instantiating field elements from integer values, we need to take care of wrap-around in integer operations, as this might yield inconsistent behavior. For instance, `FieldElement::from(1)-FieldElement::from(2) != FieldElement::from(1-2)`, as -1 in the field is a different element than -1 in `u64` (i.e., `u64::MAX`). The same issue applies to the `FieldElement.createFromLong()` Java API, which relies on `FieldElement::from` under the hood.

This PR fixes the misuses of such conversion found in the library and adds documentation to `FieldElement.createFromLong()` Java method to specify that the function safely converts only positive integer values.

### DIFF
--- a/api/src/cctp_calls.rs
+++ b/api/src/cctp_calls.rs
@@ -217,7 +217,7 @@ fn get_naive_threshold_sig_circuit_prover_data(
     }
 
     //Compute b as v-t and convert it to field element
-    let b = FieldElement::from(valid_signatures - threshold);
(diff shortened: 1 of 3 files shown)


mikelodder7 (commented):
Simple enough

DanieleDiBenedetto (approved):

PR #102, opened 2023-01-09
NCC-NLU-Low-Inadequate Validator Keys Merkle Tree

mikelodder7:
Just the non-circuit based code. Once we agree on some parameters and methods then I can add the circuit based logic and promote to PR

### DIFF
--- a/demo-circuit/src/naive_threshold_sig_w_key_rotation/data_structures.rs
+++ b/demo-circuit/src/naive_threshold_sig_w_key_rotation/data_structures.rs
@@ -9,7 +9,7 @@ use primitives::{
 
 use crate::{common::NULL_CONST, Error};
 
-const VALIDATOR_HASH_SALT: u8 = 0u8;
+pub const VALIDATOR_HASH_SALT: u8 = 0u8;
 
 //TODO: It would be nice using a constant generic here
 #[derive(Clone)]
--- a/demo-circuit/src/naive_threshold_sig_w_key_rotation/data_structures.rs
+++ b/demo-circuit/src/naive_threshold_sig_w_key_rotation/data_structures.rs
@@ -10,7 +10,6 @@ use primitives::{
 use crate::{common::NULL_CONST, Error};
 
 pub const VALIDATOR_HASH_SALT: u8 = 0u8;
-pub const INITIAL_EPOCH_ID: u32 = 0;
 
 //TODO: It would be nice using a constant generic here
 #[derive(Clone)]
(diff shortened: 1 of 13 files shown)


jni/src/main/java/com/horizen/schnorrnative/ValidatorKeysUpdatesList.java
```diff
@@ -80,24 +91,35 @@ public SchnorrPublicKey[] getMasterKeys() {
...
+    }
+
     /**
     * Utility: computes the validator keys merkle root given the set of signing keys
     * and master keys passed as input to the function.
     *
     * @param signingKeys - The signers' public keys.
     * @param masterKeys - The signers' master public keys.
     * @param maxPks - Maximum number of pks
+    * @param withdrawalCertificate - The withdrawal certificate tied to the root hash
```
nicholas-mainardi, line 109: Documentation seems to be inconsistent with the function signature

jni/src/main/java/com/horizen/schnorrnative/ValidatorKeysUpdatesList.java
```diff
@@ -5,6 +5,7 @@
 
 import com.horizen.librustsidechains.Library;
 import com.horizen.librustsidechains.FieldElement;
+import com.horizen.certnative.WithdrawalCertificate;
```
nicholas-mainardi, line 8: Is this import really necessary?

demo-circuit/src/naive_threshold_sig_w_key_rotation/tests/utils.rs
```diff
@@ -85,12 +89,25 @@ pub(crate) fn cert_to_msg(withdrawal_certificate: &WithdrawalCertificateData) ->
     .unwrap()
 }
 
-pub(crate) fn updated_key_msg(pk: FieldBasedSchnorrPk<G2Projective>) -> FieldElement {
+pub(crate) fn updated_key_msg(
```
nicholas-mainardi, line 92: This method could probably be moved to `ValidatorKeysUpdates`, since this logic is also needed when computing signatures in the primitive

demo-circuit/src/naive_threshold_sig_w_key_rotation/data_structures.rs
```diff
@@ -9,6 +9,9 @@ use primitives::{
 
 use crate::{common::NULL_CONST, Error};
 
+pub const VALIDATOR_HASH_SALT: u8 = 0u8;
```
DanieleDiBenedetto, line 12: I'm not fully comfortable with this, as hardcoding it inside the circuit means that if and when we change the protocol, we need also to change the circuit and keep the backward compatibility with the previous one(s).
It's also true that if we change the protocol, most likely also the circuit changes so perhaps I'm overthinking it.

demo-circuit/src/naive_threshold_sig_w_key_rotation/data_structures.rs
```diff
@@ -9,6 +9,9 @@ use primitives::{
...
+pub const INITIAL_EPOCH_ID: u32 = 0;
```
DanieleDiBenedetto, line 13: I'm not fully comfortable in putting a meaningful value for something that is supposed to be dummy. It's true that we rely on the MC being honest and putting the correct dummy value, but I'm also thinking about some possible replay attacks for the first certificates.
But maybe, again, I'm overthinking it

PR #112, opened 2023-02-13
Add ScCommitmentCertPath.updateScCommitmentPath()

la10736:
Add ScCommitmentCertPath.updateScCommitmentPath() to handle the case where sdk can just rebuilt the sidechain tree but not the complete commitment tree.

### DIFF
--- a/api/src/lib.rs
+++ b/api/src/lib.rs
@@ -6028,9 +6028,9 @@ ffi_export!(
         path: JObject,
         sc_commitment_path: JObject,
     ) -> jboolean {
-        let path: &mut ScCommitmentCertPath = dbg!(path).as_native_ref_mut_unchecked(env);
+        let path: &mut ScCommitmentCertPath = path.as_native_ref_mut_unchecked(env);
         let sc_commitment_path: &GingerMHTPath =
-            dbg!(sc_commitment_path).as_native_ref_unchecked(env);
+            sc_commitment_path.as_native_ref_unchecked(env);
 
         match sc_commitment_path
             .clone()
--- a/api/src/lib.rs
+++ b/api/src/lib.rs
@@ -50,7 +50,7 @@ use demo_circuit::{
     type_mapping::*,
 };
 
-use jni_wrapper::{AsNativeRefMut, AsNativeRef, JNINativeWrapper};
+use jni_wrapper::{AsNativeRef, AsNativeRefMut, JNINativeWrapper};
 use primitives::{
     bytes_to_bits, signature::schnorr::field_based_schnorr::FieldBasedSchnorrPk, FieldBasedHash,
     FieldBasedMerkleTree, FieldBasedMerkleTreePath, FieldBasedSparseMerkleTree, FieldHasher,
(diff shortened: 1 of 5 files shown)


jni/src/test/java/com/horizen/commitmenttreenative/ScCommitmentCertPathTest.java
```diff
@@ -156,4 +160,90 @@ public void completeTest() {
...
+                commTreeReal.addCertLeaf(scId1Bytes, h);
+            }
+            for (byte[] h : certs2HashBytes) {
+                commTreeReal.addCertLeaf(scId2Bytes, h);
+            }
+
+            try (
+                    FieldElement expectedScTxCommitmentRoot = commTreeReal.getCommitment().get();
+                    MerklePath correctScPath = commTreeReal.getScCommitmentMerklePath(scId1Bytes).get();
+                    CommitmentTree commTreeFake = CommitmentTree.init();) {
```
paolocappelletti, line 196: can we rename commTreeFake in commTreePartial ? since it will the way it will work on sdk implementation for real :-)

paolocappelletti (approved):
looks good, just a little rename suggestion

jni/src/test/java/com/horizen/commitmenttreenative/ScCommitmentCertPathTest.java
la10736: Ok... I'll do it

jni/src/main/java/com/horizen/commitmenttreenative/ScCommitmentCertPath.java
```diff
@@ -87,4 +88,16 @@ private void check() {
...
     }
+
+    private native boolean nativeUpdateScCommitmentPath(MerklePath correctScPath) throws IllegalArgumentException;
+
+    /**
+     * Update the path from the sidechain tx root to the commitment root
+     * 
+     * @param path The new merkele path from the sdichain root to the commitment root
+     * @throws IllegalArgumentException if the given merkle path is not valid (wrong lenght)
+     */
```
DanieleDiBenedetto, line 99:
```suggestion
     * @param path The new merkle path from the sidechain root to the commitment root
     * @throws IllegalArgumentException if the given merkle path is not valid (wrong length)
     */
```

api/src/lib.rs
```diff
@@ -6024,6 +6022,32 @@ ffi_export!(
...
 
+ffi_export!(
+    fn Java_com_horizen_commitmenttreenative_ScCommitmentCertPath_nativeUpdateScCommitmentPath(
+        env: JNIEnv,
+        path: JObject,
+        sc_commitment_path: JObject,
+    ) -> jboolean {
+        let path: &mut ScCommitmentCertPath = dbg!(path).as_native_ref_mut_unchecked(env);
+        let sc_commitment_path: &GingerMHTPath =
+            dbg!(sc_commitment_path).as_native_ref_unchecked(env);
```
DanieleDiBenedetto, line 6033: Do we need these `dbg!` ?

DanieleDiBenedetto (approved):

api/src/lib.rs
la10736: :eyes:

PR #115, opened 2023-02-15
add seed length checks

mikelodder7:


DanieleDiBenedetto:
After a brief discussion with @i-Alex we decided to close this and give the SDK the responsibility to call this function with an appropriate seed length

PR #117, opened 2023-02-23
Update vrf parameters

nicholas-mainardi:
This PR introduces the changes to the VRF parameters and methods necessary after the changes introduced in ginger-lib with [PR 196 ](https://github.com/HorizenOfficial/ginger-lib/pull/196) and [PR 200](https://github.com/HorizenOfficial/ginger-lib/pull/200)

### DIFF
--- a/demo-circuit/src/constants/mod.rs
+++ b/demo-circuit/src/constants/mod.rs
@@ -1,4 +1,4 @@
-use algebra::{biginteger::BigInteger256 as BigInteger, field_new, Field, ProjectiveCurve};
+use algebra::{biginteger::BigInteger256 as BigInteger, field_new, Field};
 
 use primitives::{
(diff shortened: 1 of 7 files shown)


demo-circuit/src/constants/mod.rs
```diff
@@ -61,7 +61,7 @@ impl NaiveThresholdSigParams {
 #[derive(Clone)]
 pub struct VRFWindow {}
 impl PedersenWindow for VRFWindow {
-    const WINDOW_SIZE: usize = 128;
+    const WINDOW_SIZE: usize = 63;
```
DanieleDiBenedetto, line 64: I think we can bring it down to 43

DanieleDiBenedetto (changes requested):

DanieleDiBenedetto (approved):

